PR #2, opened 2026-01-12
nushell: init wrapper + zoxide integration

Squawkykaka:
This pr adds a  nushell wrapper. it also adds a integration with zoxide in order to use zoxides features.

### DIFF
--- a/modules/nushell.nix
+++ b/modules/nushell.nix
@@ -1,0 +1,108 @@
+{ adios }:
+let
+  inherit (adios) types;
+in
+{
+  name = "nushell";
+
+  inputs = {
+    nixpkgs.path = "/nixpkgs";
+    mkWrapper.path = "/mkWrapper";
+  };
+
+  options = {
+    config = {
+      type = types.str;
+      description = ''
+        Config to be injected into the wrapped package's `config.nu`.
+
+        See the nushell documentation for valid options:
+        https://www.nushell.sh/book/configuration.html
+
+        Disjoint with the `configFile` option.
+      '';
+      mutatorType = types.str;
+      mergeFunc =
+        { mutators, options }:
+        let
+          inherit (builtins) attrValues concatStringsSep;
+        in
+        concatStringsSep "\n" (attrValues mutators);
+    };
+    configFile = {
+      type = types.pathLike;
+      description = ''
+        `config.nu` file to be injected into the wrapped package.
+
+        See the nushell documentation on file syntax:
+        https://www.nushell.sh/book/configuration.html
+
+        Disjoint with the `config` option.
+      '';
+    };
+
+    environment = {
+      type = types.attrs;
+      description = ''
+        Attrset of environment variables to be injected into the
+        wrapped packages `config.nu`.
+
+        See the nushell documentation for valid options:
+        https://www.nushell.sh/book/configuration.html
+
+        Example:
+        ```
+        {
+          HELLO = "test";
+          FOO = "3";
+        }
+        ```
+
+        Maps to:
+        ```
+        $env.HELLO = "test"
+        $env.FOO = "3"
+
+        Disjoint with the `environmentFile` option.
+        ```
+      '';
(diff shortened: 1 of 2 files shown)


llakala:
Be warned, I'm gonna force push and fixup some small things.

llakala:
Force pushed to fix formatting and basic style. Now I can get into the real review.

modules/nushell.nix
```diff
@@ -0,0 +1,63 @@
...
+
+  inputs = {
+    nixpkgs.path = "/nixpkgs";
+    mkWrapper.path = "/mkWrapper";
+  };
+
+  options = {
+    config = {
+      type = types.string;
+      description = "This config maps to the config.nu file";
```
llakala: These descriptions should match the style used in the other modules. Check some of them out and try to use the style seen there.

modules/nushell.nix
```diff
@@ -0,0 +1,63 @@
...
+      mutatorType = types.string;
+      mergeFunc =
+        { mutators, options }:
+        let
+          inherit (builtins) attrValues concatStringsSep;
+        in
+        concatStringsSep "\n" (attrValues mutators);
+    };
+    configFile = {
```
llakala: Needs a description.

modules/nushell.nix
```diff
@@ -0,0 +1,63 @@
...
+      type = types.pathLike;
+    };
+
+    environment = {
+      type = types.attrs;
+      description = "This takes an attrset, and converts it to the nushell environment syntax.";
```
llakala: See above note on description style guides

modules/nushell.nix
```diff
@@ -0,0 +1,63 @@
...
+      mutatorType = types.attrs;
+      mergeFunc = adios.lib.mergeFuncs.mergeAttrsRecursively;
+    };
```
llakala, line 71: Needs description.

modules/nushell.nix
```diff
@@ -0,0 +1,61 @@
...
+    environmentFile = {
+      type = types.pathLike;
+    };
+  };
+
+  impl =
+    { options, inputs }:
+    let
+      inherit (inputs.nixpkgs) pkgs lib;
```
llakala: I'd prefer if we inherited direct values from `pkgs` and `lib`, as is standard practice in the other modules.

modules/zoxide.nix
```diff
@@ -45,6 +45,52 @@ in {
       ${lib.getExe finalWrapper} init fish ${concatStringsSep " " options.flags} | source
     '';
 
+  mutations."/nushell".config =
+    { inputs, options }:
+    let
+      finalWrapper = options {};
+      inherit (inputs.nixpkgs) lib;
+    in
+    ''
```
llakala: This is a _lot_ of code, and I don't have the nushell knowledge to judge its quality. I think it may be best to save this for a future PR.

modules/zoxide.nix
```diff
@@ -45,6 +45,52 @@ in {
...
+          }
+        }
+        cd $path
+      }
+
+      def --env --wrapped __zoxide_zi [...rest:string] {
+        cd $'(${lib.getExe finalWrapper} query --interactive -- ...$rest | str trim -r -c "\n")'
+      }
+
+      alias z = __zoxide_z
```
llakala: If we did include this, these should be configurable.

llakala (changes requested):

modules/zoxide.nix
Squawkykaka: This code is autogenerated by zoxide, i just included it and used derivation paths instead

llakala: Ah, then can't we just do:
```
  ${lib.getExe finalWrapper} init nushell | source
```

That's what we do for fish. I'm not sure what the nushell `source` equivalent is, but the basic idea works.

llakala:
Thanks!

PR #3, opened 2026-01-15
Add eza wrapper

aabush64:
Adds module with support for setting default flags and theme.yml (via path or attrset).

### DIFF
--- a/docs/options.json
+++ b/docs/options.json
@@ -21,6 +21,25 @@
     }
   },
 
+  "eza": {
+    "flags": {
+      "description": "Flags to be appended by default when running eza.",
+      "type": "listOf<string>"
+    },
+    "package": {
+      "description": "The eza package to be wrapped.",
+      "type": "derivation"
+    },
+    "themeConfig": {
+      "description": "Settings to be injected into the wrapped package's `theme.yml`. See `https://github.com/eza-community/eza/blob/main/man/eza_colors-explanation.5.md` for valid options Disjoint with the `themeFile` option.",
+      "type": "attrs"
+    },
+    "themeFile": {
+      "description": "`theme.yml` file to be injected into the wrapped package. See `https://github.com/eza-community/eza/blob/main/man/eza_colors-explanation.5.md` for valid options Disjoint with the `themeConfig` option.",
+      "type": "pathLike"
+    }
+  },
+
   "firefox": {
     "autoConfigFiles": {
       "description": "`autoconfig.js` files to be injected into the wrapped package. See the Firefox documentation: https://support.mozilla.org/en-US/kb/customizing-firefox-using-autoconfig",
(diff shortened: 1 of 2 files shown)


docs/options.json
```diff
@@ -13,14 +13,30 @@
       "type": "derivation"
     }
   },
-
```
llakala, line 16: Somehow, all the newlines have been stripped out of this file.

modules/eza.nix
```diff
@@ -0,0 +1,78 @@
...
+        Flags to be appended by default when running eza.
+      '';
+    };
+
+    themeFile = {
+      type = types.pathLike;
+      description = ''
+        `theme.yml` file to be injected into the wrapped package.
+
+        See `eza_colors-explanation(5)` for valid options
```
llakala: Is there a link we can post instead? That's what I've been doing elsewhere (pointing to man7).

modules/eza.nix
```diff
@@ -0,0 +1,78 @@
...
+
+        Disjoint with `themeConfig` option.
```
llakala: Can you add a "the" here, to match the standard? See contrib docs. Same for the themeConfig option.

modules/eza.nix
```diff
@@ -0,0 +1,78 @@
...
+
+  options = {
+    flags = {
+      type = types.listOf types.string;
+      description = ''
+        Flags to be appended by default when running eza.
+      '';
+    };
+
+    themeFile = {
```
llakala: I like putting file options below their RFC42 equivalents. I also think there should be no whitespace between two "friend options" - so the themeConfig and themeFile wouldn't have a newline between them.

modules/eza.nix
```diff
@@ -0,0 +1,78 @@
...
+    }:
+    let
+      inherit (inputs.nixpkgs.pkgs) writeText;
+      inherit (inputs.nixpkgs.lib.generators) toJSON;
+    in
+    assert !(options ? themeConfig && options ? themeFile);
+    inputs.mkWrapper {
+      inherit (options) package flags;
+      preWrap = ''
+        mkdir -p $out/config
```
llakala: Could we change this name to something other than "config"? I would probably name it something like /eza, but maybe /eza-config is better.

modules/eza.nix
```diff
@@ -0,0 +1,78 @@
...
+        "$out/config/theme.yml" =
+          if options ? themeFile then
+            options.themeFile
+          else if options ? themeConfig then
+            (writeText "theme" (toJSON options.themeConfig))
+          else
+            null;
+      };
+      environment = {
+        EZA_CONFIG_HOME = "$out/config/";
```
llakala: Nitpick - no trailing slash here.

llakala (changes requested):
Generally looks great - you did a really good job following style guidelines. I'm working on the infra for a configFile option right now.

llakala:
Oh, also please rebase and change the commit message to something like `modules/eza: init`.

llakala:
Force pushed to rebase on main and fix some style nits. Thanks!

PR #4, opened 2026-01-15
helix: init wrapper

Squawkykaka:
This wrapper adds helix.
currently only issue might be formatting & i did see 1 issue about "too many symlinks"

### DIFF
--- a/modules/helix.nix
+++ b/modules/helix.nix
@@ -1,0 +1,153 @@
+{ adios }:
+let
+  inherit (adios) types;
+in
+{
+  name = "helix";
+
+  inputs = {
+    nixpkgs.path = "/nixpkgs";
+    mkWrapper.path = "/mkWrapper";
+  };
+
+  options = {
+    settings = {
+      type = types.attrs;
+      description = ''
+        Settings to be injected into the wrapped package's `config.toml`.
+
+        See the documentation:
+        https://docs.helix-editor.com/configuration.html
+
+        Disjoint with the `configFile` option.
+      '';
+    };
+    settingsFile = {
+      type = types.pathLike;
+      description = ''
+        `config.toml` file to be injected into the wrapped package.
+
+        See the documentation for valid options:
+        https://docs.helix-editor.com/configuration.html
+
+        Disjoint with the `config` option.
+      '';
+    };
+    themes = {
+      type = types.attrsOf types.attrs;
+      description = ''
+        An attrset of custom themes, mapping the name of a theme to its settings.
+
+        See the documentation for valid options:
+        https://docs.helix-editor.com/themes.html
+
+        Disjoint with the `themeDir` option.
+      '';
+    };
+    themeDir = {
+      type = types.pathLike;
+      description = ''
+        Folder containing theme configuration files to be injected into the wrapped package.
+
+        This folder should contain one or multiple `$theme_name.toml` files.
+
+        Disjoint with the `themes` option.
+      '';
+    };
+    languages = {
+      type = types.attrs;
+      description = ''
+        Languages to be injected into the wrapped package's `languages.toml`.
+
+        See the documentation for valid options:
+        https://docs.helix-editor.com/languages.html
+
+        Disjoint with the `languagesFile` option.
+      '';
+    };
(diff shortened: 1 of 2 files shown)


modules/helix.nix
```diff
@@ -0,0 +1,150 @@
...
+        See the documentation on valid options:
+        https://docs.helix-editor.com/languages.html
+
+        Disjoint with the `languages` option.
+      '';
+    };
+    extraPackages = {
+      type = types.listOf types.derivation;
+      description = ''
+        Packages to be automatically added to helix's path. Mainly used to language servers.
```
es-sai-fi:
```suggestion
        Packages to be automatically added to helix's path. Mainly used for language servers and formatters.
```

Squawkykaka:
@es-sai-fi fixed

modules/helix.nix
```diff
@@ -0,0 +1,150 @@
...
+{
+  name = "helix";
+
+  inputs = {
+    nixpkgs.path = "/nixpkgs";
+    mkWrapper.path = "/mkWrapper";
+  };
+
+  options = {
+    config = {
```
llakala: I'd prefer this to be called `settings` - hm does the same.

modules/helix.nix
llakala: This comment was resolved, but seems to have not been fixed?

modules/helix.nix
```diff
@@ -0,0 +1,150 @@
...
+            )
+          )
+        else
+          {};
+    in
+    assert !(options ? config && options ? configFile);
+    assert !(options ? themes && options ? themeDir);
+    assert !(options ? languages && options ? languageDir);
+    inputs.mkWrapper {
+      name = "helix";
```
llakala: Is this necessary? The name should be inferred properly.

modules/helix.nix
```diff
@@ -0,0 +1,150 @@
...
+        "$out/helix/config.toml" =
+          if options ? configFile then
+            options.configFile
+          else if options ? config then
+            tomlGenerator.generate "config.toml" options.config
+          else
+            null;
+        "$out/helix/languages.toml" =
+          if options ? languages then
+            options ? languages
```
llakala: You mean `options.languages`.

modules/helix.nix
```diff
@@ -0,0 +1,150 @@
...
+        See the documentation for valid options:
+        https://docs.helix-editor.com/configuration.html
+
+        Disjoint with the `config` option.
+      '';
+    };
+    themes = {
+      type = types.attrsOf types.attrs;
+      description = ''
+        `themename.toml` file to be injected into the wrapped package.
```
llakala: This description is inaccurate. It's an attribute set of themes, with each name mapping to the theme's contents. I'd prefer this to be clarified in the description. I also believe the type is wrong - (it should be `types.attrsOf types.string`).

llakala (changes requested):
There's a few bugs here from options that I believe you didn't test, or you would've gotten an error. For future PRs, it would help if you could try setting all the options to ensure they function.

modules/helix.nix
```diff
@@ -0,0 +1,149 @@
...
+{
+  name = "helix";
+
+  inputs = {
+    nixpkgs.path = "/nixpkgs";
+    mkWrapper.path = "/mkWrapper";
+  };
+
+  options = {
+settings = {
```
llakala: Formatting broken. Please try to check these things.

Squawkykaka: sorry about that :(

modules/helix.nix
```diff
@@ -0,0 +1,149 @@
...
+        See the documentation on valid options:
+        https://docs.helix-editor.com/languages.html
+
+        Disjoint with the `languages` option.
+      '';
+    };
+    extraPackages = {
+      type = types.listOf types.derivation;
+      description = ''
+        Packages to be automatically added to helix's path. Mainly used to language servers and formatters.
```
llakala: Missing the word "set" here.

modules/helix.nix
llakala: All good, just grumpy.

PR #5, opened 2026-01-16
modules/waybar: init

aabush64:
Adds module for Waybar by Alexays.

### DIFF
--- a/modules/waybar.nix
+++ b/modules/waybar.nix
@@ -1,0 +1,115 @@
+{ adios }:
+let
+  inherit (adios) types;
+in
+{
+  name = "waybar";
+
+  inputs = {
+    mkWrapper.path = "/mkWrapper";
+    nixpkgs.path = "/nixpkgs";
+  };
+
+  options = {
+    settings = {
+      type = types.attrs;
+      description = ''
+        Settings to be injected into the wrapped package's `config.jsonc`.
+
+        See the documentation for valid options:
+        https://github.com/Alexays/Waybar/wiki/Configuration
+
+        Disjoint with the `configFile` option.
+      '';
+    };
+    configFile = {
+      type = types.pathLike;
+      description = ''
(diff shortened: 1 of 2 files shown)


modules/waybar.nix
```diff
@@ -0,0 +1,129 @@
...
+        Live reloading of themes can be accomplished with an impure path and options.interactiveEnv set to true.
+
+        Disjoint with the `barStyle` option.
+      '';
+    };
+
+    interactiveEnv = {
+      type = types.bool;
+      description = ''
+        Set GTK_DEBUG=interactive to launch the wrapped package with the GTK CSS Inspector.
```
llakala: Change this to "sets"?

modules/waybar.nix
```diff
@@ -0,0 +1,129 @@
...
+
+        See the documentation for use of the inspection tool:
+        https://developer.gnome.org/documentation/tools/inspector.html
+
+        Can be used with an impure path in options.cssFile to enable live theme reloading.
+      '';
+    };
+
+    barID = {
```
llakala: Like I said on discord, I think we can just remove this and have the user set it in their config.

modules/waybar.nix
```diff
@@ -0,0 +1,129 @@
...
+      description = ''
+        CSS to be injected into the wrapped package's `style.css`.
+
+        See the documentation for writing waybar themes:
+        https://github.com/Alexays/Waybar/wiki/Styling
+
+        Disjoint with the `cssFile` option.
+      '';
+    };
+    cssFile = {
```
llakala, line 48: Maybe a `css` option too that sets the contents of the css file, which we can then do a `pkgs.writeText` on? I enjoy being ridiculous when it comes to alternatives.

llakala (changes requested):
Looks good other than a few small things. Thanks as always for your well-thought out and readable contributions!

PR #6, opened 2026-01-18
modules/fuzzel: init

aabush64:
Creates a new wrapper for fuzzel, a rofi/dmenu alternative by dnkl

### DIFF
--- a/modules/fuzzel.nix
+++ b/modules/fuzzel.nix
@@ -1,0 +1,154 @@
+{ adios }:
+let
+  inherit (adios) types;
+in {
+  name = "fuzzel";
+
+  inputs = {
+    mkWrapper.path = "/mkWrapper";
+    nixpkgs.path = "/nixpkgs";
+  };
+
+  options = {
+    settings = {
+      type = types.attrs;
+      description = ''
+        Settings to be injected into the wrapped package's `fuzzel.ini`.
+
+        See the documentation for valid options:
+        https://codeberg.org/dnkl/fuzzel/src/branch/master/doc/fuzzel.ini.5.scd
+
+        Disjoint with the `configFile` option.
+      '';
+    };
+    configFile = {
+      type = types.pathLike;
+      description = ''
+        `fuzzel.ini` file to be injected into the wrapped package.
+
+        See the documentation for syntax and valid options:
+        https://codeberg.org/dnkl/fuzzel/src/branch/master/doc/fuzzel.ini.5.scd
+
+        Disjoint with the `settings` option.
+      '';
+    };
+
+    dmenuFlags = {
+      type =
+        (types.struct "dmenuFlags" {
+          dmenu = types.bool;
+          dmenu0 = types.bool;
+          withNth = types.string;
+          acceptNth = types.string;
+          delimitNth = types.string;
+          noEmpty = types.bool;
+        }).override
+          {
+            total = false;
+            unknown = false;
+          };
+      description = ''
+        Dmenu related flags to pass to fuzzel when using `settings`/`configFile`
+
+        See the documentation for valid syntax and formatting of the related flags:
+        https://codeberg.org/dnkl/fuzzel/src/branch/master/doc/fuzzel.1.scd
+
+        Requires that either `settings`/`configFile` be set.
+      '';
+    };
+
+    logFlags = {
+      type =
+        (types.struct "logFlags" {
+          logLevel = types.string;
+          colorize = types.string;
+          noSyslog = types.bool;
+          printTiming = types.bool;
+        }).override
+          {
+            total = false;
+            unknown = false;
+          };
+      description = ''
+        Logging related flags to pass to fuzzel when using `settings`/`configFile`
+
+        See the documentation for valid syntax and formatting of the related flags:
+        https://codeberg.org/dnkl/fuzzel/src/branch/master/doc/fuzzel.1.scd
+
+        Requires that either `settings`/`configFile` be set.
+      '';
+    };
+
+    package = {
+      type = types.derivation;
+      description = "The fuzzel package to be wrapped.";
+      defaultFunc = { inputs }: inputs.nixpkgs.pkgs.fuzzel;
+    };
+  };
+
+  impl =
+    { options, inputs }:
+    let
+      inherit (inputs.nixpkgs.pkgs) writeText;
+      inherit (inputs.nixpkgs.lib) optionals;
+      inherit (inputs.nixpkgs.lib.generators) toIni;
+      inherit (builtins) head concatLists;
+    in
+    assert (
+      if options ? dmenuFlags then
+        (options ? settings || options ? configFile)
+      else
+        true
+    );
+    assert (
+      if options ? logFlags then
+        (options ? settings || options ? configFile)
+      else
+        true
+    );
+    assert !(options ? settings && options ? configFile);
+    let
+      dmenuFlags =
+        if options ? dmenuFlags then
+          optionals (options.dmenuFlags ? dmenu && options.dmenuFlags.dmenu) [ "--dmenu" ]
+          ++ optionals (options.dmenuFlags ? dmenu0 && options.dmenuFlags.dmenu0) [ "--dmenu0" ]
+          ++ optionals (options.dmenuFlags ? withNth) [ "--with-nth=${options.dmenuFlags.withNth}" ]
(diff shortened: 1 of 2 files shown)


modules/fuzzel.nix
```diff
@@ -0,0 +1,93 @@
...
+      description = ''
+        Settings to be injected into the wrapped package's `fuzzel.ini`.
+
+        See the documentation for valid options:
+        https://codeberg.org/dnkl/fuzzel/src/branch/master/doc/fuzzel.ini.5.scd
+        
+        Disjoint with the `configIni` option.
+      '';
+    };
+    configIni = {
```
llakala: Rename to `configFile`.

modules/fuzzel.nix
```diff
@@ -0,0 +1,93 @@
...
+      defaultFunc = { inputs }: inputs.nixpkgs.pkgs.fuzzel;
+    };
+  };
+
+  impl =
+    { options, inputs }:
+    let
+      inherit (inputs.nixpkgs.pkgs) writeText;
+      inherit (inputs.nixpkgs.lib.generators) toIni;
+      inherit (inputs.nixpkgs.lib.lists) last mutuallyExclusive;
```
llakala: `builtins.head` can do `lib.last` better. These are singleton lists, so identical.

modules/fuzzel.nix
```diff
@@ -0,0 +1,93 @@
...
+  options = {
+    flags = {
+      type = types.listOf types.string;
+      description = ''
+        Flags to be appended by default when running fuzzel.
+
+        See the documentation for valid options:
+        https://codeberg.org/dnkl/fuzzel/src/branch/master/doc/fuzzel.1.scd
+
+        *Not* disjoint with `settings/configIni` but conflicting settings between the configuration and set flags may have unexpected behavior.
```
llakala: Interesting. I need to think more about this. Is there anything that can _only_ be done in flags, and can't be done in settings / settingsFile?

llakala (changes requested):

modules/fuzzel.nix
```diff
@@ -0,0 +1,93 @@
...
+          [ ];
+      flags =
+        if options ? flags then
+          assert (mutuallyExclusive options.flags configFlag);
+          options.flags ++ configFlag
+        else
+          configFlag;
+    in
+    inputs.mkWrapper {
+      name = "fuzzel";
```
llakala, line 143: Is this necessary? Might work without it.

modules/fuzzel.nix
```diff
@@ -0,0 +1,93 @@
...
+      inherit (options) package;
+      inherit flags;
+      preWrap =
+      if options ? settings || options ? configIni then
+      ''
+        exec $out/bin/fuzzel --check-config ${last configFlag}
```
llakala: This doesn't seem right to me... this is the buildtime code, not the runtime code. The runtime code that's generated is actually C that's compiled to an executable (let me know if you want a demo of how to preview this). As it stands, I believe this would execute fennel during the build. Maybe you want a conditional on `flags` instead?

llakala: Or I see, maybe check-config is fine? Because it's just validating the config works. IDK. Maybe `pkgs.runCommandLocal` makes more sense for this?

modules/fuzzel.nix
aabush64: There are multiple flags, mostly related to dmenu compatibility (and some logging) that are not configurable via the settings/settingsFile. You can find those flags mainly towards the end of the [man page found here](https://codeberg.org/dnkl/fuzzel/src/branch/master/doc/fuzzel.1.scd).

modules/fuzzel.nix
aabush64: Could you reference what "this" is referring to?

modules/fuzzel.nix
aabush64: Yes this is just validating whether the given configuration is valid. Looking into `runCommandLocal`, would generating an extra derivation be overkill for a validation step?

modules/fuzzel.nix
llakala: Sorry - `name` field here

modules/fuzzel.nix
llakala: Good point. Have you tested this current version to see if it validates properly? If so, I'm fine keeping it, but I want some guarantee.

aabush64: Yes, I gave it both valid and invalid configurations, an invalid configuration causes a build failure via non-zero exit code which is expected behavior.

llakala: Okay, we can stick with this version, then. My original comment stemmed from a misunderstanding.

modules/fuzzel.nix
llakala: Gotcha. Let me think about this one a bit.

modules/fuzzel.nix
aabush64: Ah, now that i think of it, I'm not sure why it was added! oops >.>

llakala:
Force pushed to fix formatting. Maybe I broke the formatting check, I'll revert my recent commit.

modules/fuzzel.nix
```diff
@@ -0,0 +1,212 @@
...
+    { options, inputs }:
+    let
+      inherit (inputs.nixpkgs.pkgs) writeText;
+      inherit (inputs.nixpkgs.lib.generators) toIni;
+      inherit (inputs.nixpkgs.lib.lists) flatten;
+      inherit (builtins) head;
+    in
+    assert !(options ? flags && options ? settings);
+    assert !(options ? flags && options ? configFile);
+    assert (options ? dmenuFlags -> (options ? settings || options ? configFile));
```
llakala: Rewrite to `if then else`

modules/fuzzel.nix
```diff
@@ -0,0 +1,212 @@
...
+                []
+            )
+            (
+              if options.dmenuFlags ? delimitNth then
+                [ "--nth-delimiter=${options.dmenuFlags.delimitNth}" ]
+              else
+                []
+            )
+            (
+              if options.dmenuFlags ? noEmpty then
```
llakala: Need an && to check that this is true.

modules/fuzzel.nix
```diff
@@ -0,0 +1,212 @@
...
+in {
+  name = "fuzzel";
+
+  inputs = {
+    mkWrapper.path = "/mkWrapper";
+    nixpkgs.path = "/nixpkgs";
+  };
+
+  options = {
+    flags = {
```
llakala: I'd prefer if we removed the flags options altogether, in favor of the structs.

modules/fuzzel.nix
```diff
@@ -0,0 +1,212 @@
...
+          [ "--config=${options.configFile}" ]
+        else if options ? settings then
+          [ "--config=${writeText "fuzzel.ini" (toIni options.settings)}" ]
+        else
+          [];
+      flags =
+        if options ? flags then
+          options.flags
+        else
+          flatten (dmenu ++ logs ++ configFlag);
```
llakala: Use `builtins.concatLists` here instead of `lib.flatten`.

llakala (changes requested):
They don't let me review without a comment in here!

aabush64:
Once again I forget the options!

llakala:
CI failing on formatting.

modules/fuzzel.nix
```diff
@@ -0,0 +1,203 @@
...
+      if options ? logFlags then
+        (options ? settings || options ? configFile)
+      else
+        true
+    );
+    assert !(options ? settings && options ? configFile);
+    let
+      dmenu =
+        if options ? dmenuFlags then
+          [
```
llakala: I believe this is a rewrite from the previous version. I would be fine with this if the formatter made it look decent, but it's horrifically ugly. I would personally write it to look similarly to [the swaybg module](https://github.com/llakala/adios-wrappers/blob/d511cf7a3fd3daaf10cbf44e58dc7025989a8327/modules/swaybg.nix#L57-L73).

llakala (changes requested):
Good other than this.

llakala:
Going to force push to do what I have in mind with the flags, since it's easier than explaining it.

llakala:
Force pushed. I'll wait for your approval on the final result before merging. I just got rid of the concatLists.

aabush64:
No objections, good to go.

llakala:
Thanks! Sorry for the delay on this PR.

PR #8, opened 2026-01-20
modules/discordo: init

poach3r:
This provides a wrapper for the
[discordo](https://github.com/ayn2op/discordo). In a few days I'll likely create a single pr with the rest of my wrappers but I'd like to put this forward for feedback.

### DIFF
--- a/modules/discordo.nix
+++ b/modules/discordo.nix
@@ -1,0 +1,90 @@
+{adios}: let
+  inherit (adios) types;
+in {
+  name = "discordo";
+
+  inputs = {
+    mkWrapper.path = "/mkWrapper";
+    nixpkgs.path = "/nixpkgs";
+  };
+
+  options = {
+    settings = {
+      type = types.attrs;
+      description = ''
+        Settings to be injected into the wrapped package's 'config.toml'.
+
+        See the default configuration for valid options:
+        https://github.com/ayn2op/discordo/blob/main/internal/config/config.toml
+
+        Disjoint with the `configFile` option.
+      '';
+    };
+    configFile = {
+      type = types.pathLike;
+      description = ''
+        `config.toml` file to be injected into the wrapped package.
+
+        See the default configuration for valid options:
+        https://github.com/ayn2op/discordo/blob/main/internal/config/config.toml
+
+        Disjoint with the `settings` option.
+      '';
+    };
+
+    tokenFile = {
+      type = types.pathLike;
+      description = ''
+        A file containing a discord token which is read at runtime and injected into the wrapped package's environment.
+
+        I recommend using a solution like [sops-nix](https://github.com/Mic92/sops-nix) or [agenix](https://github.com/ryantm/agenix) to encrypt your token.
+      '';
+    };
+
+    flags = {
+      type = types.listOf types.string;
+      default = [];
+      description = ''
+        Flags to be automatically appended when running discordo.
+      '';
+    };
+
+    package = {
+      type = types.derivation;
+      defaultFunc = {inputs}: inputs.nixpkgs.pkgs.discordo;
+      description = "The discordo package to be wrapped.";
+    };
+  };
+
+  impl = {
+    options,
+    inputs,
+  }: let
+    generator = inputs.nixpkgs.pkgs.formats.toml {};
+  in
+    assert !(options ? settings && options ? configFile);
+      inputs.mkWrapper {
+        inherit (options) package flags;
+        preWrap = ''
+          mkdir -p $out/discordo/
+        '';
+        environment = {
+          XDG_CONFIG_HOME = "$out";
+          DISCORDO_TOKEN =
+            if options ? tokenFile
+            then {
+              value = options.tokenFile;
+              readFromFile = true;
+            }
+            else null;
+        };
+        symlinks = {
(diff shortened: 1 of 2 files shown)


modules/discordo.nix
```diff
@@ -0,0 +1,90 @@
...
+        `config.toml` to be injected into the wrapped package.
+
+        See the default configuration for valid options:
+        https://github.com/ayn2op/discordo/blob/main/internal/config/config.toml
+
+        Disjoint with the `settings` option.
+      '';
+    };
+
+    token = {
```
llakala: Rename this to `tokenFile`.

modules/discordo.nix
```diff
@@ -0,0 +1,90 @@
...
+      description = ''
+        Settings to be injected into the wrapped package's 'config.toml'.
+
+        See the default configuration for valid options:
+        https://github.com/ayn2op/discordo/blob/main/internal/config/config.toml
+
+        Disjoint with the `configFile` option.
+      '';
+    };
+
```
llakala: We typically have no whitespace between options that are disjoint.

modules/discordo.nix
```diff
@@ -0,0 +1,90 @@
...
+{
+  name = "discordo";
+
+  inputs = {
+    nixpkgs.path = "/nixpkgs";
+    mkWrapper.path = "/mkWrapper";
+  };
+
+  options = {
+    package = {
```
llakala: Package options typically come at the end of the `options` block.

modules/discordo.nix
```diff
@@ -0,0 +1,90 @@
...
+    token = {
+      type = types.path;
+      description = ''
+        A file containing a discord token which is read at runtime and injected into the wrapped package's environment.
+
+        I recommend using a solution like [sops-nix](https://github.com/Mic92/sops-nix) or [agenix](https://github.com/ryantm/agenix) to encrypt your token.
+      '';
+    };
+
+    flags = {
```
llakala, line 44: This is fine for now, but in the future, we should have an impure variant of this kind of thing. However, that requires me to do some more work with my fork of `makeBinaryWrapper`, so it can support reading flags from a file at runtime and appending them. TLDR ignore this, just noting that it's something that I plan to fix.

modules/discordo.nix
```diff
@@ -0,0 +1,90 @@
...
+      options,
+      inputs,
+    }:
+    let
+      generator = inputs.nixpkgs.pkgs.formats.toml { };
+    in
+    assert !(options ? settings && options ? configFile);
+    inputs.mkWrapper {
+      inherit (options) package flags;
+      name = "discordo";
```
llakala: I believe this name is unnecessary. It defaults to the pname of the package, which should be fine here.

modules/discordo.nix
```diff
@@ -0,0 +1,90 @@
...
+      preWrap = ''
+        mkdir -p $out/discordo/
+      '';
+      environment = {
```
llakala: Typically we put environment below symlinks.

modules/discordo.nix
```diff
@@ -0,0 +1,90 @@
...
+        XDG_CONFIG_HOME = "$out";
+        DISCORDO_TOKEN = if options ? token then "$(cat ${options.token})" else "";
```
llakala: I'm unsure on whether this will work for mkWrapper, which uses makeBinaryWrapper instead of just makeWrapper. Have you tested this, and if not, can you?

modules/discordo.nix
```diff
@@ -0,0 +1,91 @@
...
+      };
+      symlinks = {
+        "$out/discordo/config.toml" =
+          if options ? configFile then
+            options.configFile
+          else if options ? settings then
+            (generator.generate "config.toml" options.settings)
```
llakala: Parentheses unnecessary here.

modules/discordo.nix
```diff
@@ -0,0 +1,91 @@
...
+      description = ''
+        A file containing a discord token which is read at runtime and injected into the wrapped package's environment.
+
+        I recommend using a solution like [sops-nix](https://github.com/Mic92/sops-nix) or [agenix](https://github.com/ryantm/agenix) to encrypt your token.
+      '';
+    };
+
+    flags = {
+      type = types.listOf types.string;
+			default = [];
```
llakala: Broken formatting.

llakala (changes requested):
Docs are failing CI. Run `/dev/generate-docs.sh > docs/options.json` to regenerate them.

modules/discordo.nix
```diff
@@ -0,0 +1,91 @@
...
+        https://github.com/ayn2op/discordo/blob/main/internal/config/config.toml
+
+        Disjoint with the `configFile` option.
+      '';
+    };
+
+    configFile = {
+      type = types.pathLike;
+      description = ''
+        `config.toml` to be injected into the wrapped package.
```
llakala: "`config.toml` file" is the standard for these descriptions, as documented [here](https://github.com/llakala/adios-wrappers/blob/2b7d9468919297813107f2aa9701ea34776eb206/docs/contributing.md?plain=1#L90).

llakala:
Thanks!

PR #9, opened 2026-01-21
wrappers/yazi: add flavors, theme and themeFile opts

es-sai-fi:
## Checklist

- [ x ] If I added/changed any options, I ran `./dev/generate-docs.sh > docs/options.json`
- [ x ] I checked [contributing.md](../docs/contributing.md) and my changes conform to it
- [ x ] I tested my wrapper to make sure it functioned

### DIFF
--- a/modules/yazi.nix
+++ b/modules/yazi.nix
@@ -56,6 +56,29 @@ in {
       '';
     };
 
+    theme = {
+      type = types.attrs;
+      description = ''
+        Theme settings to be injected into the wrapped package's `theme.toml`.
+
+        See the documentation for valid options:
+        https://yazi-rs.github.io/docs/configuration/theme/
+
+        Disjoint with the `themeFile` option.
+      '';
+    };
+    themeFile = {
+      type = types.pathLike;
+      description = ''
+        `theme.toml` file to be injected into the wrapped package.
+
+        See the documentation for valid options:
+        https://yazi-rs.github.io/docs/configuration/theme/
+
+        Disjoint with the `theme` option.
+      '';
+    };
+
     initLua = {
       type = types.string;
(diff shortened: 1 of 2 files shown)


modules/yazi.nix
```diff
@@ -111,6 +134,14 @@ in {
         Each attribute should map the name of a plugin (suffixed with `.yazi`) to the path or derivation containing the plugin's contents.
       '';
     };
+    flavors = {
+      type = types.attrsOf types.pathLike;
+      description = ''
+        Attribute set of flavors to be injected into the wrapped package.
+
+        Each attribute should map the name of a plugin (suffixed with `.yazi`) to the path or derivation containing the flavor's contents.
```
llakala:
```suggestion
        Each attribute should map the name of a flavor (suffixed with `.yazi`) to the path or derivation containing the flavor's contents.
```

llakala (changes requested):

modules/yazi.nix
```diff
@@ -56,6 +56,29 @@ in {
...
+        See the documentation for valid options:
+        https://yazi-rs.github.io/docs/configuration/theme/
+
+        Disjoint with the `themeFile` option.
+      '';
+    };
+    themeFile = {
+      type = types.pathLike;
+      description = ''
+        `theme.toml` file to injected into the wrapped package.
```
llakala: Oops, missed on first review: "to be injected".

llakala (changes requested):
Looks good other than one thing which I missed on first pass.

modules/yazi.nix
llakala: Actually, I can fix this myself.

PR #10, opened 2026-01-21
modules/alacritty: init

es-sai-fi:
## Checklist

- [ x ] If I added/changed any options, I ran `./dev/generate-docs.sh > docs/options.json`
- [ x ] I checked [contributing.md](../docs/contributing.md) and my changes conform to it
- [ x ] I tested my wrapper to make sure it functioned

### DIFF
--- a/modules/alacritty.nix
+++ b/modules/alacritty.nix
@@ -1,0 +1,64 @@
+{adios}:
+let
+  inherit (adios) types;
+in {
+  name = "alacritty";
+
+  inputs = {
+    mkWrapper.path = "/mkWrapper";
+    nixpkgs.path = "/nixpkgs";
+  };
+
+  options = {
+    settings = {
+      type = types.attrs;
+      description = ''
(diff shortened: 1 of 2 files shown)


llakala:
No thank you, copilot.

modules/alacritty.nix
```diff
@@ -0,0 +1,64 @@
...
+
+        Disjoint with the `configFile` option.
+      '';
+    };
+    configFile = {
+      type = types.pathLike;
+      description = ''
+        `alacritty.toml` file to be injected into the wrapped package.
+
+        See the kitty documentation:
```
llakala: Still says "kitty"

llakala (changes requested):
Please disable Copilot for future PRs.

llakala (approved):
We don't have a standard formatter yet, so merging this for now (I'll do a treewide when we get one), but we prefer a more nixfmt-like style, so use that for future modules if you can.

PR #12, opened 2026-01-21
wrappers/zsh: init

poach3r:
## Checklist

### DIFF
--- a/modules/zsh.nix
+++ b/modules/zsh.nix
@@ -1,0 +1,124 @@
+# TODO integration with zoxide, plugins, etc
+{ adios }:
+let
+  inherit (adios) types;
+in {
+  name = "zsh";
+
+  inputs = {
+    mkWrapper.path = "/mkWrapper";
+    nixpkgs.path = "/nixpkgs";
+  };
+
+  options = {
+    zshrc = {
+      type = types.string;
+      description = ''
+        zsh script to be injected into the wrapped package's `.zshrc`.
+
+        Disjoint with the `zshrcFiles` option.
+      '';
+      mutatorType = types.string;
+      mergeFunc =
+        { mutators, options }:
+        let
+          inherit (builtins) attrValues concatStringsSep;
+        in
+        concatStringsSep "\n" (attrValues mutators);
+    };
+    extraZshrc = {
+      type = types.string;
+      description = ''
+        zsh script to be injected at the end of the wrapped package's `.zshrc`.
+
+        Disjoint with the `zshrcFiles` option.
+      '';
+      mutatorType = types.string;
+      mergeFunc =
+        { mutators, options }:
+        let
+          inherit (builtins) attrValues concatStringsSep;
+        in
+        concatStringsSep "\n" (attrValues mutators);
+    };
+    zshrcFiles = {
+      type = types.listOf types.pathLike;
+      description = ''
+        zsh files to be sourced in the wrapped package's `.zshrc` file.
+
+        Disjoint with the `zshrc` option.
+      '';
+      mutatorType = types.listOf types.pathLike;
+      mergeFunc = adios.lib.mergeFuncs.concatLists;
+    };
+    extraZshrcFiles = {
+      type = types.listOf types.pathLike;
+      description = ''
+        zsh files to be sourced at the end of the wrapped package's `.zshrc` file.
+
+        Disjoint with the `zshrc` option.:
+      '';
+      mutatorType = types.listOf types.pathLike;
+      mergeFunc = adios.lib.mergeFuncs.concatLists;
+    };
+
+    extraPackages = {
+      type = types.listOf types.derivation;
+      description = ''
+        Runtime dependencies to be injected into the wrapped package's path.
+      '';
+      mutatorType = types.listOf types.pathLike;
+      mergeFunc = adios.lib.mergeFuncs.concatLists;
+    };
+
+    package = {
+      type = types.derivation;
+      defaultFunc = { inputs }: inputs.nixpkgs.pkgs.zsh;
+      description = "The zsh package to be wrapped.";
+    };
+  };
+
+  impl =
+    { options, inputs }:
+    let
+      inherit (inputs.nixpkgs.pkgs) writeText;
+      inherit (inputs.nixpkgs.lib) makeBinPath;
+      inherit (builtins) concatStringsSep;
+      sourceFiles = files: (concatStringsSep "\n" (map (file: "source ${file}") files));
+      zshrc =
+        if options ? zshrcFiles then
+          (sourceFiles options.zshrcFiles)
+          + (
+            if options ? extraZshrcFiles then
+              "\n" + (sourceFiles options.extraZshrcFiles)
+            else
+              ""
+          )
+        else if options ? zshrc then
+          options.zshrc
+          + (
+            if options ? extraZshrc then
+              "\n" + options.extraZshrc
+            else
+              ""
+          )
+        else
+          null;
+    in
+    assert !(options ? zshrc && options ? zshrcFiles);
+    # extraZshrcFiles requires zshrcFiles to be set.
+    assert (if options ? extraZshrcFiles && !(options ? zshrcFiles) then false else true);
+    # extraZshrc requires zshrc to be set.
+    assert (if options ? extraZshrc && !(options ? zshrc) then false else true);
+    inputs.mkWrapper {
+      inherit (options) package;
+      wrapperArgs =
+        if options ? extraPackages then "--prefix PATH : ${makeBinPath options.extraPackages}" else null;
+      symlinks = {
+        "$out/.zshrc" = if options ? zshrc || options ? zshrcFile then writeText ".zshrc" zshrc else null;
+      };
+      environment = {
+        ZDOTDIR = if options ? zshrc || options ? zshrcFile then "$out" else null;
+      };
+    };
+}
--- a/docs/options.json
+++ b/docs/options.json
@@ -568,5 +568,37 @@
       "description": "The zoxide package to be wrapped.",
       "type": "derivation"
     }
+  },
+
+  "zsh": {
+    "extraPackages": {
+      "description": "Runtime dependencies to be injected into the wrapped package's path.",
+      "mutatorType": "listOf<pathLike>",
+      "type": "listOf<derivation>"
+    },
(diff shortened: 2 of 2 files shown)


modules/zsh.nix
```diff
@@ -0,0 +1,136 @@
...
+  options = {
+    zshrc = {
+      type = types.string;
+      description = ''
+        zsh to be injected into the wrapped package's `.zshrc`.
+
+        Disjoint with the `zshrcFiles` option.
+      '';
+      mutatorType = types.string;
+      defaultFunc = { inputs }: inputs.nixpkgs.pkgs.callPackage ../pkgs/zsh/zshrc.nix { };
```
llakala: Left this in.

modules/zsh.nix
```diff
@@ -0,0 +1,136 @@
...
+  inputs = {
+    mkWrapper.path = "/mkWrapper";
+    nixpkgs.path = "/nixpkgs";
+  };
+
+  options = {
+    zshrc = {
+      type = types.string;
+      description = ''
+        zsh to be injected into the wrapped package's `.zshrc`.
```
llakala: Missing word.

modules/zsh.nix
```diff
@@ -0,0 +1,136 @@
...
+        }:
+        let
+          inherit (builtins) attrValues concatStringsSep;
+        in
+        concatStringsSep "\n" (attrValues mutators);
+    };
+    extraZshrc = {
+      type = types.string;
+      description = ''
+        zsh to be injected at the end of the wrapped package's `.zshrc`.
```
llakala: Another missing word.

modules/zsh.nix
```diff
@@ -0,0 +1,136 @@
...
+      '';
+      mutatorType = types.listOf types.pathLike;
+      mergeFunc = adios.lib.mergeFuncs.concatLists;
+    };
+
+    package = {
+      type = types.derivation;
+      defaultFunc = { inputs }: inputs.nixpkgs.pkgs.zsh;
+      description = ''
+        The zsh package to be wrapped.
```
llakala: This doesn't need a multiiline striing.

modules/zsh.nix
```diff
@@ -0,0 +1,136 @@
...
+  };
+
+  impl =
+    {
+      options,
+      inputs,
+    }:
+    let
+      inherit (inputs.nixpkgs) pkgs lib;
+
```
llakala: The amount of newlines here is stressing me out.

modules/zsh.nix
```diff
@@ -0,0 +1,136 @@
...
+        if options ? zshrcFiles then
+          (sourceFiles options.zshrcFiles)
+          + (if options ? extraZshrcFiles then "\n" + (sourceFiles options.extraZshrcFiles) else "")
+        else if options ? zshrc then
+          options.zshrc + (if options ? extraZshrc then "\n" + options.extraZshrc else "")
+        else
+          "";
+    in
+    assert !(options ? zshrc && options ? zshrcFiles);
+    assert (if options ? extraZshrcFiles && !(options ? zshrcFiles) then false else true);
```
llakala, line 110: This is better, but can we have the comments back too?

modules/zsh.nix
```diff
@@ -0,0 +1,136 @@
...
+    };
+  };
+
+  impl =
+    {
+      options,
+      inputs,
+    }:
+    let
+      inherit (inputs.nixpkgs) pkgs lib;
```
llakala: I'd prefer inheriting values directly from pkgs and lib.

modules/zsh.nix
```diff
@@ -0,0 +1,136 @@
...
+
+      sourceFiles = files: (builtins.concatStringsSep "\n" (map (file: "source ${file}") files));
```
llakala: Inherit builtins where possible.

modules/zsh.nix
```diff
@@ -0,0 +1,136 @@
...
+
+      zshrc =
+        if options ? zshrcFiles then
+          (sourceFiles options.zshrcFiles)
+          + (if options ? extraZshrcFiles then "\n" + (sourceFiles options.extraZshrcFiles) else "")
+        else if options ? zshrc then
+          options.zshrc + (if options ? extraZshrc then "\n" + options.extraZshrc else "")
+        else
+          "";
```
llakala: Null by default - mkWrapper ignores anything null.

modules/zsh.nix
```diff
@@ -0,0 +1,136 @@
...
+
+      symlinks = {
+        "$out/.zshrc" = pkgs.writeText ".zshrc" zshrc;
+      };
+
+      environment =
+        if (options ? zshrc || options ? zshrcFile) then
+          {
+            ZDOTDIR = "$out";
```
llakala: environment ignores anything null too - do something like `ZDOTDIR = if options ? zshrc || options ? zshrcFile then "$out" else null;`

modules/zsh.nix
```diff
@@ -0,0 +1,136 @@
...
+          options.zshrc + (if options ? extraZshrc then "\n" + options.extraZshrc else "")
+        else
+          "";
+    in
+    assert !(options ? zshrc && options ? zshrcFiles);
+    assert (if options ? extraZshrcFiles && !(options ? zshrcFiles) then false else true);
+    assert (if options ? extraZshrc && !(options ? zshrc) then false else true);
+    inputs.mkWrapper {
+      inherit (options) package;
+
```
llakala: The newlines! My eyes!

modules/zsh.nix
```diff
@@ -0,0 +1,136 @@
...
+      inherit (inputs.nixpkgs) pkgs lib;
+
+      sourceFiles = files: (builtins.concatStringsSep "\n" (map (file: "source ${file}") files));
+
+      zshrc =
+        if options ? zshrcFiles then
+          (sourceFiles options.zshrcFiles)
+          + (if options ? extraZshrcFiles then "\n" + (sourceFiles options.extraZshrcFiles) else "")
+        else if options ? zshrc then
+          options.zshrc + (if options ? extraZshrc then "\n" + options.extraZshrc else "")
```
llakala: Inherit builtins directly too.

llakala (changes requested):
Generally, the descriptions don't really follow the recommended phrasing as describes [here](https://github.com/llakala/adios-wrappers/blob/5a79ebfc0d27528d00054904a1dbcf2acfcba9e5/docs/contributing.md?plain=1#L65) - would appreciate you taking another look at them.

modules/zsh.nix
poach3r: > Null by default - mkWrapper ignores anything null.

This can't be null as its fed into `pkgs.writeText`. I could check if either `zshrc` or `zshrcFiles` exist and set `$out/.zshrc` to null otherwise if you'd like. Personally, I think the way it works currently is fine since anyone using the wrapper should be using either `zshrc` or `zshrcFiles`, but I know you dislike these edge cases.

poach3r:
Not sure if this is intended behavior or not but it seems `makeWrapper`'s `wrapperArgs` doesn't accept null values? I'm reverting it to be an empty string if `options.extraPackages` isn't set.

llakala:
> Not sure if this is intended behavior or not but it seems `makeWrapper`'s `wrapperArgs` doesn't accept null values? I'm reverting it to be an empty string if `options.extraPackages` isn't set.

Just added this.

modules/zsh.nix
llakala: My dislike for the edge cases is just aesthetic. You already have the same condition for the `environment`, so I think it'd be nice to add the condition to the symlinks as well, so no zshrc gets generated if neither are set.

llakala (changes requested):
Going to force push for a formatting tweak (multiline `{ foo, bar }` expressions can remain on one line if you start them like that). I might enforce that for a certain number of elements.

llakala:
Ah, I can't force push to your branch since it's on main. Here's the patch:
```patch
diff --git a/modules/zsh.nix b/modules/zsh.nix
index e3e3f15..41480d2 100644
--- a/modules/zsh.nix
+++ b/modules/zsh.nix
@@ -20,10 +20,7 @@ in {
       '';
       mutatorType = types.string;
       mergeFunc =
-        {
-          mutators,
-          options,
-        }:
+        { mutators, options }:
         let
           inherit (builtins) attrValues concatStringsSep;
         in
@@ -38,10 +35,7 @@ in {
       '';
       mutatorType = types.string;
       mergeFunc =
-        {
-          mutators,
-          options,
-        }:
+        { mutators, options }:
         let
           inherit (builtins) attrValues concatStringsSep;
         in
@@ -85,10 +79,7 @@ in {
   };
 
   impl =
-    {
-      options,
-      inputs,
-    }:
+    { options, inputs }:
     let
       inherit (inputs.nixpkgs.lib) makeBinPath;
```

modules/zsh.nix
```diff
@@ -0,0 +1,124 @@
...
+    # extraZshrcFiles requires zshrcFiles to be set.
+    assert (if options ? extraZshrcFiles && !(options ? zshrcFiles) then false else true);
+    # extraZshrc requires zshrc to be set.
+    assert (if options ? extraZshrc && !(options ? zshrc) then false else true);
+    inputs.mkWrapper {
+      inherit (options) package;
+      wrapperArgs =
+        if options ? extraPackages then "--prefix PATH : ${makeBinPath options.extraPackages}" else null;
+      symlinks = {
+        "$out/.zshrc" = if (options ? zshrc || options ? zshrcFile) then writeText ".zshrc" zshrc else null;
```
llakala: Parentheses aren't needed for the condition here.

modules/zsh.nix
```diff
@@ -0,0 +1,124 @@
...
+      };
+      environment = {
+        ZDOTDIR = if (options ? zshrc || options ? zshrcFile) then "$out" else null;
```
llakala: Same here.

modules/zsh.nix
```diff
@@ -0,0 +1,124 @@
...
+        else if options ? zshrc then
+          options.zshrc
+          + (
+            if options ? extraZshrc then
+              "\n" + options.extraZshrc
+            else
+              ""
+          )
+        else
+          "";
```
llakala: This can now be null if you bump the adios-wrappers fork.

llakala (approved):

PR #15, opened 2026-01-21
wrappers/bottom: init

es-sai-fi:
## Checklist

### DIFF
--- a/modules/bottom.nix
+++ b/modules/bottom.nix
@@ -1,0 +1,69 @@
+{ adios }:
+let
+  inherit (adios) types;
+in
+{
+  name = "bottom";
+
+  inputs = {
+    mkWrapper.path = "/mkWrapper";
+    nixpkgs.path = "/nixpkgs";
+  };
+
+  options = {
+    settings = {
+      type = types.attrs;
+      description = ''
+        Settings to be injected into the wrapped package's `bottom.toml`.
+
+        See the bottom documentation for valid options:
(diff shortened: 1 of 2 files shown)


es-sai-fi:
I need feedback on the usage of flags since I don't think it's intended to be used this way. Problem I found is something like "-C $out/bottom/bottom.toml" won't work because $out doesn't expand (i think), I tried also "-C '$out/bottom/bottom.toml'" and "-C \"$out/bottom/bottom.toml\""... We could also set XDG_CONFIG_HOME to $out.

es-sai-fi:
I did some more testing and yeah couldn't get it to work unless I separate the string...

modules/bottom.nix
```diff
@@ -0,0 +1,55 @@
...
+      description = "The bottom package to be wrapped.";
+      defaultFunc = { inputs }: inputs.nixpkgs.pkgs.bottom;
+    };
+  };
+
+  impl =
+    { options, inputs }:
+    let
+      generator = inputs.nixpkgs.pkgs.formats.toml {};
+      generatedConfig = options.configFile or (generator.generate "bottom.toml" options.settings);
```
llakala: The reason we typically do symlinks injected into the derivation is for easier debugging. I would keep a symlink inside the derivation, and set XDG_CONFIG_HOME.

llakala (changes requested):

llakala (approved):
This is my favorite wrapper you've written - all very clear. Great job!

PR #16, opened 2026-01-21
wrappers/alacritty: refactor impl

es-sai-fi:
simplify wrapping, there's no need of making a dir and a symlink for something that can be done with just a flag that points to the generated config in the nix store

## Checklist

### DIFF
--- a/modules/alacritty.nix
+++ b/modules/alacritty.nix
@@ -44,21 +44,11 @@ in {
     { options, inputs }:
     let
       generator = inputs.nixpkgs.pkgs.formats.toml {};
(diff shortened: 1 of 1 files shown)


llakala:
We don't use `or` syntax for a specific reason - we want the wrapper to behave expectedly if neither option is set. Can you make this an `if then else` statement to be consistent, and add a `throw` explaining that one of the options must be set?

llakala:
Actually, I think I prefer the symlink method for debugging purposes, since you can inspect the contents of the derivation and immediately tell what config is being injected. If you can measure a meaningful performance difference between the two, I'll accept this, but I don't expect that's the case.

PR #18, opened 2026-01-21
wrappers/tealdeer: init

es-sai-fi:
## Checklist

### DIFF
--- a/modules/tealdeer.nix
+++ b/modules/tealdeer.nix
@@ -1,0 +1,67 @@
+{adios}:
+let
+  inherit (adios) types;
+in {
+  name = "tealdeer";
+
+  inputs = {
+    mkWrapper.path = "/mkWrapper";
+    nixpkgs.path = "/nixpkgs";
+  };
+
+  options = {
+    settings = {
+      type = types.attrs;
+      description = ''
+        Settings to be injected into the wrapped package's `config.toml`.
+
+        See the tealdeer documentation:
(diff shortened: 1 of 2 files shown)


modules/tealdeer.nix
```diff
@@ -0,0 +1,65 @@
...
+      '';
+      symlinks = {
+        "$out/tealdeer/config.toml" =
+          if options ? configFile
+            then options.configFile
+          else if options ? settings
+            then generator.generate "config.toml" options.settings
+          else null;
+      };
+      environment.TEALDEER_CONFIG_DIR = "$out/tealdeer";
```
llakala: We never do `.` syntax here, let's be consistent across modules.

modules/tealdeer.nix
```diff
@@ -0,0 +1,65 @@
...
+    { options, inputs }:
+    let
+      generator = inputs.nixpkgs.pkgs.formats.toml {};
+    in
+    assert !(options ? settings && options ? configFile);
+    inputs.mkWrapper {
+      inherit (options) package;
+      binaryPath = "$out/bin/tldr";
+      preWrap = ''
+        mkdir -p $out/tealdeer
```
llakala: I'd name the folder something like `tealdeer-config`.

llakala (changes requested):

llakala (approved):
For future PRs, please use nixfmt instead of alejandra.

PR #20, opened 2026-01-22
wrappers/zellij: init

es-sai-fi:
## Checklist

### DIFF
--- a/modules/zellij.nix
+++ b/modules/zellij.nix
@@ -1,0 +1,137 @@
+{ adios }:
+let
+  inherit (adios) types;
+in {
+  name = "zellij";
+
+  inputs = {
+    mkWrapper.path = "/mkWrapper";
+    nixpkgs.path = "/nixpkgs";
+  };
+
+  # The reason for there not being RFC42 options like `settings` or `layouts` is
+  # because KDL is xml-like meaning there's no one way of it being encoded to
+  # something like Nix.
+  options = {
+    configContents = {
+      type = types.string;
+      description = ''
+        Config to be injected into the wrapped package's `config.kdl`.
+
+        See the documentation for valid options:
+        https://zellij.dev/documentation/configuration.html
+
+        Disjoint with the `configFile` option.
+      '';
+    };
+    configFile = {
+      type = types.pathLike;
+      description = ''
+        `config.kdl` file to be injected into the wrapped package.
+
+        See the documentation for valid options:
+        https://zellij.dev/documentation/configuration.html
+
+        Disjoint with the `configContents` option.
+      '';
+    };
+
+    layoutsContents = {
+      type = types.attrsOf types.str;
+      description = ''
+        Custom layouts to be injected into the wrapped package.
+
+        See the documentation to learn how to make custom layouts and use them:
+        https://zellij.dev/documentation/layouts.html
+
+        Disjoint with the `layoutsFiles` option.
+      '';
+    };
+    layoutsFiles = {
+      type = types.listOf types.pathLike;
+      description = ''
+        Files containing custom layouts to be injected into the wrapped package.
+
+        See the documentation to learn how to make custom layouts and use them:
+        https://zellij.dev/documentation/layouts.html
+
+        Disjoint with the `layoutsContents` option.
+      '';
+    };
+
+    themes = {
+      type = types.attrsOf types.pathLike;
+      description = ''
+        Attribute set of themes to be injected into the wrapped package.
+
+        See the zellij documentation to learn how to make custom themes and use them:
+        https://zellij.dev/documentation/themes.html
+
+        Each attribute name is used as the theme name. The corresponding value must be
+        a folder or derivation whose top level contains a file named `<name>.kdl`.
+
+        For example, an entry `themes.dracula = <drv>;` where `<drv>` contains
+        `dracula.kdl` will result in `themes/dracula.kdl`.
+      '';
+    };
+
+    package = {
+      type = types.derivation;
+      description = "The zellij package to be wrapped.";
+      defaultFunc = { inputs }: inputs.nixpkgs.pkgs.zellij;
+    };
+  };
+
+  impl =
+    { options, inputs }:
+    let
+      inherit (builtins) listToAttrs attrNames;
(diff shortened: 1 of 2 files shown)


modules/zellij.nix
```diff
@@ -0,0 +1,91 @@
...
+        https://zellij.dev/documentation/themes.html
+
+        Each attribute name is used as the theme name. The corresponding value must be
+        a path or derivation whose top level contains a file named `<name>.kdl`, which
+        will be installed as `themes/<name>.kdl`.
+
+        For example, an entry `themes.dracula = <drv>;` where `<drv>` contains
+        `dracula.kdl` will result in `themes/dracula.kdl`.
+      '';
+    };
```
llakala, line 76: Newline here is our standard practice.

modules/zellij.nix
```diff
@@ -0,0 +1,91 @@
...
+    themes = {
+      type = types.attrsOf types.pathLike;
+      description = ''
+        Attribute set of themes to be injected into the wrapped package.
+
+        See the zellij documentation to learn how to make custom themes and use them:
+        https://zellij.dev/documentation/themes.html
+
+        Each attribute name is used as the theme name. The corresponding value must be
+        a path or derivation whose top level contains a file named `<name>.kdl`, which
```
llakala:
```suggestion
        a folder or derivation whose top level contains a file named `<name>.kdl`, which
```

modules/zellij.nix
```diff
@@ -0,0 +1,91 @@
...
+  inherit (adios) types;
+in {
+  name = "alacritty";
+
+  inputs = {
+    mkWrapper.path = "/mkWrapper";
+    nixpkgs.path = "/nixpkgs";
+  };
+
+  options = {
```
llakala, line 15: I'd add a comment here explaining that there are no rfc42 options because KDL is xml-like and has no one way of being encoded in something like Nix.

modules/zellij.nix
llakala: I also think you can remove the "which will be installed at" part of the sentence. I think that's obvious.

modules/zellij.nix
```diff
@@ -0,0 +1,91 @@
...
+        Attribute set of themes to be injected into the wrapped package.
+
+        See the zellij documentation to learn how to make custom themes and use them:
+        https://zellij.dev/documentation/themes.html
+
+        Each attribute name is used as the theme name. The corresponding value must be
+        a path or derivation whose top level contains a file named `<name>.kdl`, which
+        will be installed as `themes/<name>.kdl`.
+
+        For example, an entry `themes.dracula = <drv>;` where `<drv>` contains
```
llakala, line 73: Adios modules do have an example field now - you could put that there.

llakala: But would mean you wouldn't be able to add context about what the example would result in. Up to you.

modules/zellij.nix
```diff
@@ -0,0 +1,91 @@
+{ adios }:
+let
+  inherit (adios) types;
+in {
+  name = "alacritty";
```
llakala: Oops.

llakala (changes requested):
What beautiful formatting! Whoever made this formatter deserves a raise.

modules/zellij.nix
es-sai-fi: I'm not going to lie I almost copied and pasted your brief explanation, let me know if it's not enough and I'll come up with something else.

modules/zellij.nix
es-sai-fi: I believe it's better to leave the "example" in the desc because there's no zellij themes in nixpkgs as of now so it would be confusing to reference something that doesn't exist:

```nix
example = {
  dracula = pkgs."dracula.zellij";
  dracula = pkgs.zellijPlugins.dracula;
};
```

Let me know if the new description is better.

modules/zellij.nix
es-sai-fi: Assuming I understood correctly what you meant to say I think I fixed this, let me know.

llakala: Indeed you did.

modules/zellij.nix
llakala: Yeah, this is fine.

modules/zellij.nix
```diff
@@ -0,0 +1,94 @@
...
+            name = "$out/zellij-config/layouts/${baseNameOf path}";
+            value = path;
+          }) options.layoutsFiles
+        )
+      )
+      // optionalAttrs (options ? themes) (
+        listToAttrs (
+          map (name: {
+            name = "$out/zellij-config/themes/${name}.kdl";
+            value = "${options.themes.${name}}/${name}.kdl";
```
llakala:
```suggestion
            value = "${name}/${name}.kdl";
```

llakala (changes requested):
Will merge after the one thing I noticed.

es-sai-fi:
Should a configStr option be made?

es-sai-fi:
I made both a configStr and layoutsStrs options, I personally would use these instead of the configFiles and layoutsFiles ones because this way I don't have to make zellij have it's own folder (zellij/{default.nix,config.kdl,my-layout.kdl}), I don't know if that makes sense. I tested the new opts, both work fine.

llakala:
> I made both a configStr and layoutsStrs options, I personally would use these instead of the configFiles and layoutsFiles ones because this way I don't have to make zellij have it's own folder (zellij/{default.nix,config.kdl,my-layout.kdl}), I don't know if that makes sense. I tested the new opts, both work fine.

I'd prefer naming them `configContents` and `layoutContents`.

es-sai-fi:
> I'd prefer naming them `configContents` and `layoutContents`.

Isn't `layoutsContents` better in this case? Since it's multiple layouts:

```nix
layoutsContents = {
  my-layout = ''
    ...
  '';
  
  my-other-layout = ''
    ...
  '';
};
```

Now I'm not a native english speaker so let me know if `layoutContents` is the correct way.

llakala:
Sure, layoutsContents seems better. A struct form might be nice in the future, but we can do this for now.

PR #22, opened 2026-01-22
wrappers/yazi: fix missing dependency

es-sai-fi:
## Checklist

### DIFF
--- a/modules/yazi.nix
+++ b/modules/yazi.nix
@@ -109,6 +109,10 @@ in {
 
         This defaults to the optionalDeps of the Yazi package in nixpkgs, set here:
         https://github.com/NixOS/nixpkgs/blob/master/pkgs/by-name/ya/yazi/package.nix#L8
+
+        The dependency `File` is added regardless of the content of this option, because it's non-optional.
+        See the Yazi docs on this topic:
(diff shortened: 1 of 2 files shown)


modules/yazi.nix
```diff
@@ -109,6 +109,8 @@ in {
 
         This defaults to the optionalDeps of the Yazi package in nixpkgs, set here:
         https://github.com/NixOS/nixpkgs/blob/master/pkgs/by-name/ya/yazi/package.nix#L8
+
+        The dependency `File` is added regardless of the content of this option because it's not optional according to Yazi's docs.
```
llakala: Link to the relevant part of the docs?

llakala (changes requested):

modules/yazi.nix
es-sai-fi: Let me know if just linking them is enough, I'd do something like

```
The dependency `File` is added regardless of the content of this option because it's not optional according to Yazi's docs. For more info see:
<link>
```

But in my opinion the line is already too long.

llakala:
Force pushed for a reword of the line you changed. Will merge when CI passes.

PR #23, opened 2026-01-22
wrappers/swayidle: init

poach3r:
## Checklist

### DIFF
--- a/modules/swayidle.nix
+++ b/modules/swayidle.nix
@@ -1,0 +1,81 @@
+{ adios }:
+let
+  inherit (adios) types;
+in {
+  name = "swayidle";
+
+  inputs = {
+    mkWrapper.path = "/mkWrapper";
+    nixpkgs.path = "/nixpkgs";
+  };
+
+  options = {
+    configContents = {
+      type = types.string;
+      description = ''
+        Settings to be injected into the wrapped package's `config` file.
+
+        See the documentation for valid options: https://man.archlinux.org/man/swayidle.1
+
+        Disjoint with the `configFile` option.
+      '';
+    };
+    configFile = {
+      type = types.pathLike;
+      description = ''
+        `config` file to be injected into the wrapped package.
+
+        See the documentation for valid options: https://man.archlinux.org/man/swayidle.1
+
+        Disjoint with the `configContents` option.
+      '';
+    };
+
+    seatName = {
+      type = types.string;
+      description = "The seat name to be injected into the wrapped package's flags.";
+    };
+
+    package = {
+      type = types.derivation;
+      description = "The swayidle package to be wrapped.";
+      defaultFunc = { inputs }: inputs.nixpkgs.pkgs.swayidle;
+    };
+  };
+
+  impl =
+    { options, inputs }:
+    let
+      inherit (inputs.nixpkgs.pkgs) writeText;
+      configFlag =
+        if options ? configContent || options ? configFile then
+          [
+            "-C"
+            "$out/config"
+          ]
+        else
+          [];
+      styleFlag =
+        if options ? seat then
+          [
+            "-S"
+            options.seat
+          ]
+        else
+          [];
+    in
+    assert !(options ? configContents && options ? configFile);
+    inputs.mkWrapper {
+      inherit (options) package;
+      symlinks = {
+        "$out/config" =
+          if options ? configContents then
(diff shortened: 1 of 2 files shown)


poach3r:
I really dislike the flags block here but I couldn't figure out a better way of doing it. I'd appreciate your input.

modules/swayidle.nix
```diff
@@ -0,0 +1,85 @@
...
+  inputs = {
+    mkWrapper.path = "/mkWrapper";
+    nixpkgs.path = "/nixpkgs";
+  };
+
+  options = {
+    settings = {
+      type = types.string;
+      description = ''
+        Settings to be injected into the wrapped package's `config`.
```
llakala: Let's deviate slightly from the template, since the file just being named config makes it slightly unclear.

```suggestion
        Settings to be injected into the wrapped package's `config` file.
```

modules/swayidle.nix
```diff
@@ -0,0 +1,85 @@
...
+  name = "swayidle";
+
+  inputs = {
+    mkWrapper.path = "/mkWrapper";
+    nixpkgs.path = "/nixpkgs";
+  };
+
+  options = {
+    settings = {
+      type = types.string;
```
llakala, line 14: `settings` options are never strings. Maybe `configContents`?

modules/swayidle.nix
```diff
@@ -0,0 +1,85 @@
...
+      description = ''
+        `config` file to be injected into the wrapped package.
+
+        See the documentation for valid options: https://man.archlinux.org/man/swayidle.1
+
+        Disjoint with the `settings` option.
+      '';
+    };
+
+    seat = {
```
llakala: Maybe rename to `seatName`?

modules/swayidle.nix
```diff
@@ -0,0 +1,85 @@
...
+      inherit (inputs.nixpkgs.pkgs) writeText;
+    in
+    assert !(options ? settings && options ? configFile);
+    inputs.mkWrapper {
+      inherit (options) package;
+      symlinks = {
+        "$out/config" =
+          if options ? settings then
+            (writeText "config" options.settings)
+          else
```
llakala: I'd prefer to have the file symlinked to the derivation, even if it's a configFile. Makes debugging easier.

llakala (commented):
In regards to your question about flags, I like the way that the waybar module does it:
```nix
let
      configFlag =
        if options ? configFile then
          [ "--config=${options.configFile}" ]
        else if options ? settings then
          [ "--config=${writeText "config.jsonc" (toJSON options.settings)}" ]
        else
          [];
      styleFlag =
        if options ? cssFile then
          [ "--style=${options.cssFile}" ]
        else if options ? barStyle then
          [ "--style=${writeText "style.css" options.barStyle}" ]
        else
          [];
in
...
  flags = configFlag ++ styleFlag;
```

However, this doesn't fix the lists elements being split onto multiple lines, which is gross. I've attempted to fix this in the nixfmt fork, but haven't had much luck. I'll try to take another look at it now.

modules/swayidle.nix
poach3r: Just to clarify, configFile shouldn't be impure? Or should it still be impure but also get symlinked

llakala: Should be impure, and also get symlinked.

modules/swayidle.nix
```diff
@@ -0,0 +1,80 @@
...
+            options.seat
+          ]
+        else
+          [];
+    in
+    assert !(options ? configContents && options ? configFile);
+    inputs.mkWrapper {
+      inherit (options) package;
+      symlinks = {
+        "$out/config" = (writeText "config" options.configContents);
```
llakala: This doesn't make sense - I think you misunderstood what I meant. We should have a single flag option that points to $out/config, and then have this either point to a writeText on configContents or the configFile.

modules/swayidle.nix
```diff
@@ -0,0 +1,81 @@
...
+        else
+          [];
+    in
+    assert !(options ? configContents && options ? configFile);
+    inputs.mkWrapper {
+      inherit (options) package;
+      symlinks = {
+        "$out/config" =
+          if options ? configContents then
+            (writeText "config" options.configContents)
```
llakala: No need for parentheses here. Will force push and merge if you don't get to it tonight.

PR #26, opened 2026-01-23
flake: refactor so flake consumers don't need nixpkgs

poach3r:
This PR moves the devshell to its own `shell.nix` in `dev/` and removes the flake's `nixpkgs` input. This is accomplished by using npins to pin `nixpkgs` for the modules in `dev/`. As follows, `dev/repl.nix` now sources `nixpkgs` from npins instead of the flake's inputs.

## Checklist


poach3r:
I've thought about writing a `dev/default.nix` which sources `nixpkgs` and `adios` and exports `repl.nix` and `shell.nix` but decided against it as to not disrupt workflows. If this is something you'd like then let me know.

llakala:
Looks good in concept. Having two adios inputs between the flake and npins seems a little silly - I'll think more about it and give a proper review later today.

poach3r:
Its possible that we could use npins to source nixpkgs, then source adios from the flake's inputs. However, I find this rather clunky.

poach3r:
I forgot to say this, but would you consider adding nixd to the shell?

llakala:
> I forgot to say this, but would you consider adding nixd to the shell?

I don't use nixd, and think it's a little silly to inflate the closure size of the devshell just for people that do.

llakala:
> Its possible that we could use npins to source nixpkgs, then source adios from the flake's inputs. However, I find this rather clunky.

I find it clunky as well, but I think dev tooling should take a backseat here. In terms of repo priority, we have:
1. UX for the users (flake.nix and default.nix)
2. UX while writing modules (devshell that provides our formatter)
3. CI to test formatting and docs generation
4. UX while testing modules (repl.nix)


If we have two separate Adios inputs, I think it'll be easy to accidentally bump one and not the other. Only #3 and #4 really "require" a locally pinned version of Adios, and I'm okay with them being grosser if it means preventing bugs.

dev/repl.nix
```diff
@@ -5,9 +5,8 @@
 #
 # nix repl --expr '{ wrappers = import ./dev/repl.nix; }'
 let
-  flake = builtins.getFlake (toString ../.);
-  inherit (flake) inputs outputs;
-  pkgs = import inputs.nixpkgs {};
+  inherit (builtins.getFlake (toString ../.)) inputs outputs;
+  pkgs = import (import ./npins).nixpkgs {};
```
llakala:
```suggestion
  flake = builtins.getFlake (toString ../.);
  sources = import ./npins;
  inherit (flake) inputs outputs;
  pkgs = import sources.nixpkgs {};
```

llakala (changes requested):
Please also change docs/installation-flakes.md to remove `inputs.nixpkgs.follows`.

llakala:
Commit description is also out of date.

llakala:
Pinned to the same nixpkgs commit as before, to prevent anyone from needing to build nixfmt again. Will merge when CI passes.

llakala:
Thanks!

PR #27, opened 2026-01-23
wrappers/swaybg: init

poach3r:
## Checklist

### DIFF
--- a/modules/swaybg.nix
+++ b/modules/swaybg.nix
@@ -1,0 +1,75 @@
+{ adios }:
+let
+  inherit (adios) types;
+in {
+  name = "swaybg";
+
+  inputs = {
+    mkWrapper.path = "/mkWrapper";
+    nixpkgs.path = "/nixpkgs";
+  };
+
+  options = {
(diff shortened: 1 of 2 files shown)


llakala (changes requested):
I'd prefer if the descriptions didn't just say "passed into the ___ flag", since I think that's an implementation detail. I would say to replace it with more info explaining what each option did, but you can't really say much about what a background image does... it sets the background image. Basically strip out the implementation details and see where we can add more info about the effects of an option.

modules/swaybg.nix
```diff
@@ -0,0 +1,73 @@
...
+in {
+  name = "swaybg";
+
+  inputs = {
+    mkWrapper.path = "/mkWrapper";
+    nixpkgs.path = "/nixpkgs";
+  };
+
+  options = {
+    color = {
```
llakala, line 13: Maybe an example for this? Options support `example =` now.

llakala:
Force pushed to make the examples into simple strings, rather than an attrset.

PR #25, opened 2026-01-23
Fix typos in nushell, and fix asserts in helix

Squawkykaka:
## Checklist

- [x] If I added/changed any options, I ran `./dev/generate-docs.sh > docs/options.json`
- [x] I checked [contributing.md](../docs/contributing.md) and my changes conform to it
- [x] I tested my wrapper to make sure it functioned

PR #28, opened 2026-01-23
nushell: fix several typos, and replace env-config with string

Squawkykaka:
## Checklist
split out from [25](https://github.com/llakala/adios-wrappers/pull/25)

### DIFF
--- a/modules/nushell.nix
+++ b/modules/nushell.nix
@@ -10,10 +10,10 @@ in {
   };
 
   options = {
-    settings = {
+    shellInit = {
       type = types.string;
       description = ''
-        Config to be injected into the wrapped package's `config.nu`.
+        Shell initialisation code to be injected into the wrapped package's `config.nu`.
 
         See the nushell documentation for valid options:
         https://www.nushell.sh/book/configuration.html
@@ -35,46 +35,6 @@ in {
       '';
     };
 
-    environment = {
-      type = types.attrs;
-      description = ''
-        Attrset of environment variables to be injected into the
-        wrapped packages `config.nu`.
-
-        See the nushell documentation for valid options:
-        https://www.nushell.sh/book/configuration.html
-
-        Example:
-        ```
-        {
-          HELLO = "test";
-          FOO = "3";
-        }
-        ```
-
-        Maps to:
-        ```
-        $env.HELLO = "test"
-        $env.FOO = "3"
-        ```
-
-        Disjoint with the `environmentFile` option.
-      '';
-      mutatorType = types.attrs;
-      mergeFunc = adios.lib.merge.attrs.recursively;
-    };
-    environmentFile = {
-      type = types.pathLike;
-      description = ''
-        `env.nu` file to be injected into the wrapped package.
-
-        See the nushell documentaion on file sytax:
-        https://www.nushell.sh/book/configuration.html
-
-        Disjoint with the `environment` option.
-      '';
-    };
-
     package = {
       type = types.derivation;
       description = "The nushell package to be wrapped.";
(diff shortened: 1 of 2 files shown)


llakala:
Why make environment config into a string? I'd appreciate some motivation on a PR like this.

Squawkykaka:
I made it into a string, as the format for the env config can be different than just `$env.SOMETHING = something`, and limiting it didnt make sense

llakala:
I'd rename the option then, to better reflect its purpose. hm names it `extraEnv` - maybe that?

modules/nushell.nix
```diff
@@ -15,60 +15,55 @@ in {
...
-        $env.HELLO = "test"
-        $env.FOO = "3"
-        ```
-
         Disjoint with the `environmentFile` option.
       '';
-      mutatorType = types.attrs;
-      mergeFunc = adios.lib.mergeFuncs.mergeAttrsRecursively;
+      mutatorType = types.string;
+      mergeFunc =
```
llakala: This is equivalent to `adios.lib.mergeFuncs.concatLines`.

Squawkykaka:
Its specifically a `env.nu` file, in fact im pretty sure its not even needed anymore, i checked a issue and they say to not use it anymore

llakala:
Gotcha - removal seems fine, then.

modules/nushell.nix
```diff
@@ -28,47 +28,32 @@ in {
...
       '';
     };
 
     environment = {
-      type = types.attrs;
+      type = types.string;
       description = ''
-        Attrset of environment variables to be injected into the
-        wrapped packages `config.nu`.
+        Environment config to be injected into the wrapped package's `env.nu`.
```
llakala: I think the word `config` isn't sending the right message here. Typically config is a Nix attrset. The fish module calls inline code "shell initialization code", while the zsh module calls it "zsh script". Maybe find something more like one of these, to make it clearer what we're really talking about?

modules/nushell.nix
```diff
@@ -28,47 +28,32 @@ in {
...
 
-        See the nushell documentation for valid options:
+        See the Nushell documentation for valid options:
         https://www.nushell.sh/book/configuration.html
 
-        Example:
```
llakala, line 47: SInce you removed the inline example, adding an actual example would be nice. Something like
```nix
example = ''
  $env.HELLO = "test"
'';
```

modules/nushell.nix
```diff
@@ -15,7 +15,7 @@ in {
       description = ''
         Config to be injected into the wrapped package's `config.nu`.
```
llakala: See what I said about the word `config`.

Squawkykaka:
i did some more research and came to the conclusion that the env config is fully deprecated and only remains for backwards compat and i removed it

modules/nushell.nix
```diff
@@ -1,19 +1,20 @@
...
 
   inputs = {
-    mkWrapper.path = "/mkWrapper";
     nixpkgs.path = "/nixpkgs";
+    mkWrapper.path = "/mkWrapper";
   };
 
   options = {
-    settings = {
+    config = {
```
llakala: Now that I'm looking at it, can we rename this to something like `shellInit`?

modules/nushell.nix
```diff
@@ -23,6 +24,7 @@ in {
       mutatorType = types.string;
       mergeFunc = adios.lib.merge.strings.concatLines;
```
llakala, line 24: This line requires a rebase.

modules/nushell.nix
```diff
@@ -85,31 +47,21 @@ in {
   impl =
     { options, inputs }:
     let
-      inherit (builtins) concatStringsSep attrNames;
       inherit (inputs.nixpkgs.pkgs) writeText;
-      format =
-        attrs: concatStringsSep "\n" (map (name: "$env.${name} = \"${attrs.${name}}\"") (attrNames attrs));
```
llakala: No whitespace here.

modules/nushell.nix
```diff
@@ -23,6 +24,7 @@ in {
       mutatorType = types.string;
       mergeFunc = adios.lib.merge.strings.concatLines;
     };
+
```
llakala: No whitespace here.

modules/nushell.nix
```diff
@@ -85,31 +47,21 @@ in {
...
         else
-          [];
-      envFlag =
-        if options ? environmentFile then
-          [ "--env-config ${options.enironmentFile}" ]
-        else if options ? environment then
-          [ "--env-config ${writeText "env.nu" (format options.environment)}" ]
-        else
-          [];
+          [ ];
```
llakala: No space inside an empty list. Formatter should catch this...

modules/nushell.nix
```diff
@@ -85,31 +47,21 @@ in {
...
     in
-    assert !(options ? config && options ? configFile);
-    assert !(options ? environment && options ? environmentFile);
+    assert !(options ? settings && options ? configFile);
     inputs.mkWrapper {
+      name = "nu";
       package = options.package;
       binaryPath = "$out/bin/nu";
```
llakala: If you set the name, you don't need to set the binaryPath.

llakala (changes requested):
Sorry for all the nitpicks - keep noticing more things.

modules/nushell.nix
llakala: Oh, the formatter didn't run, I think because of the merge conflict.

llakala:
Fixed the merge conflict and the small complaints I had. Only question now is what we should name the option.

PR #32, opened 2026-02-10
modules/zsh: enable more complex nix configuration

poach3r:
This is a massive commit which does a number of things:
1. Allows for `zshrc` and `zshrcFiles` to be used with each other
2. Allows usage of `extraZshrc` and `extraZshrcFiles` without setting their corresponding options
3. Adds options for configuring aliases, variables, `setopt` options, and plugins

## Checklist

### DIFF
--- a/modules/zsh.nix
+++ b/modules/zsh.nix
@@ -176,8 +176,9 @@
       inherit (inputs.nixpkgs.lib) makeBinPath;
       inherit (builtins) concatStringsSep attrNames;
       optionalString = cond: string: if cond then string else "";
-      # Concatenates a list of strings so they're formatted properly for the .zshrc
-      concat = elems: (concatStringsSep "\n" elems) + "\n\n";
+      # Apply some function to each element in the list, then concat all the
+      # elements together with newlines in between
+      mapAndConcat = func: elems: (concatStringsSep "\n" (map func elems)) + "\n";
       # Whether a .zshrc should be generated
       shouldConfigure =
         options ? zshrc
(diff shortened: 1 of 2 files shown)


modules/zsh.nix
llakala, line 10: Would appreciate more examples, especially for options like `settings`.

modules/zsh.nix
```diff
@@ -13,38 +12,116 @@
...
+        https://zsh.sourceforge.io/Doc/Release/Options.html
+      '';
+      mutatorType = types.listOf types.pathLike;
+      mergeFunc = adios.lib.merge.lists.concat;
+    };
+
+    variables = {
+      type = types.attrs;
+      description = ''
+        Variables to be defined in the wrapped package's `.zshrc` file.
```
llakala:
```suggestion
        Environment variables to be defined in the wrapped package's `.zshrc` file.
```

modules/zsh.nix
```diff
@@ -13,38 +12,116 @@
...
       description = ''
-        zsh script to be injected at the end of the wrapped package's `.zshrc`.
-
-        Disjoint with the `zshrcFiles` option.
+        zsh script to be injected after the `zshrcFiles` option in the wrapped 
+        package's `.zshrc`.
       '';
-      mutatorType = types.string;
-      mergeFunc = adios.lib.merge.strings.concatLines;
+      mutatorType = types.listOf types.pathLike;
```
llakala: Every single option seems to have gotten the mutator type of `types.listOf types.pathLike`. This seems like a mistake.

modules/zsh.nix
```diff
@@ -70,43 +147,82 @@
...
-              "\n" + (sourceFiles options.extraZshrcFiles)
+        let
+          zshrcFiles = if options ? zshrcFiles then sourceFiles "${options.zshrcFiles}\n" else "";
+          extraZshrcFiles = if options ? extraZshrcFiles then "${options.extraZshrcFiles}\n" else "";
+          zshrcText = if options ? zshrc then "${options.zshrc}\n" else "";
+          extraZshrcText = if options ? extraZshrc then "${options.extraZshrc}\n" else "";
+          settings =
+            if options ? settings then
+              concat (
+                attrValues (mapAttrs (name: value: "${if value then "" else "un"}setopt ${name}") options.settings)
```
llakala: We never do mapAttrs in a case like this. Look at some other modules where we use listToAttrs with attrNames.

modules/zsh.nix
```diff
@@ -70,43 +147,82 @@
...
+        || options ? variables;
+      # Generates zsh sourcing all the files in a list
+      sourceFiles = files: concat (map (file: "source ${file}") files);
       zshrc =
-        if options ? zshrcFiles then
-          (sourceFiles options.zshrcFiles)
-          + (
-            if options ? extraZshrcFiles then
-              "\n" + (sourceFiles options.extraZshrcFiles)
+        let
```
llakala, line 191: A lot of these would be simplified by an `optionalString` definition. We can just define it locally rather than grabbing it from `lib`.

llakala (changes requested):
Just some things I noticed on initial review.

modules/zsh.nix
poach3r: This was an oversight caused by me copy-pasting options. Also, are there any options with a mergeFunc that shouldn't have one? I'm still hazy on where one should be used.

poach3r: Also, for options with polymorphic types (e.g `plugins`) should `mutatorType` be non-polymorphic? Alternatively, I could declare `type` in a let-in block and inherit it, then set `mutatorType` to `type`.

modules/zsh.nix
poach3r: I think you've given me this exact same criticism before, lol. Fixed and much cleaner now.

modules/zsh.nix
llakala: > This was an oversight caused by me copy-pasting options. Also, are there any options with a mergeFunc that shouldn't have one? I'm still hazy on where one should be used.

Good question. I think shells should try to have a lot of them, especially for shell init injections. I'd add them for all the things that try to inject into the zshrc.

 > Alternatively, I could declare type in a let-in block and inherit it, then set mutatorType to type.

Yep, do this (although I'd name it something clearer).

PR #35, opened 2026-03-02
fix typos, modify devshell and pull request template

es-sai-fi:
## Checklist

### DIFF
--- a/docs/installation-flakes.md
+++ b/docs/installation-flakes.md
@@ -76,4 +76,4 @@ outputs = inputs: {
 };
 ```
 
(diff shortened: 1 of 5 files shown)


llakala:
Happy to fix typos through this, and we could maybe add it as a PR check, but I'd prefer not to require people to run it themselves.

es-sai-fi:
> Happy to fix typos through this, and we could maybe add it as a PR check, but I'd prefer not to require people to run it themselves.

Alright, I'll leave that part out. I'll look into creating a new workflow or modifying an existing one to add it as a check though I gotta admit I'm not familiarized with them but well it should be easy.

es-sai-fi:
Oops, ill move that to a different pr

PR #38, opened 2026-03-20
modules/irssi: init

poach3r:
## Checklist

### DIFF
--- a/modules/irssi.nix
+++ b/modules/irssi.nix
@@ -1,0 +1,111 @@
+{ types, ... }:
+{
+  inputs = {
+    mkWrapper.path = "/mkWrapper";
+    nixpkgs.path = "/nixpkgs";
+  };
+
+  options = {
+    config = {
+      type = types.string;
+      description = ''
+        The config to be injected into the wrapped package.
+
+        See the documentation for valid options:
+        https://irssi.org/documentation/settings/
+
+        Disjoint with the `configDir` option.
+      '';
+    };
+    configDir = {
+      type = types.pathLike;
+      description = ''
+        The path of `.irssi` to be injected into the wrapped package.
+
+        See the documentation for valid options:
+        https://irssi.org/documentation/settings/
+
+        Disjoint with the `config` option.
+      '';
+    };
+
+    autoconnect = {
+      type = types.union [
+        (types.struct "autoconnectWithoutPassword" {
+          server = types.string;
+          port = types.string;
+        })
+        (types.struct "autoconnect" {
+          server = types.string;
+          password = types.string;
+          port = types.string;
+        })
+      ];
+      description = ''
+        The server to be automatically connected to.
+      '';
+      example = {
+        server = "irc.libera.chat";
+        port = "6697";
+        password = "hunter2"; # This is optional
+      };
+    };
+
+    nickname = {
+      type = types.string;
+      description = "The default nickname to be injected into the wrapped package.";
+    };
+
+    hostname = {
+      type = types.string;
+      description = "The default hostname to be injected into the wrapped package.";
+    };
+
+    package = {
+      type = types.derivation;
+      description = "The irssi package to be wrapped.";
+      defaultFunc = { inputs }: inputs.nixpkgs.pkgs.irssi;
+    };
+  };
+
+  impl =
+    { options, inputs }:
+    let
+      inherit (inputs.nixpkgs.pkgs) writeText;
+      inherit (inputs.nixpkgs.lib) optional optionals;
+      homeFlag = optional (options ? configDir) "--home=$out/irssi";
+      configFlag = optional (options ? config) "--config=$out/irssi/config";
+      autoconnectFlag = optionals (options ? autoconnect) (
+        [
+          "--connect=${options.autoconnect.server}"
+          "--port=${options.autoconnect.port}"
+        ]
+        ++ optional (options.autoconnect ? password) "--password=${options.autoconnect.password}"
+      );
+      nicknameFlag = optionals (options ? nickname) [
+        "-n"
+        options.nickname
+      ];
+      hostnameFlag = optionals (options ? hostname) [
+        "-h"
+        options.hostname
+      ];
+    in
+    assert !(options ? configDir && options ? config);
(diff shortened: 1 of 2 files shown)


modules/irssi.nix
```diff
@@ -0,0 +1,129 @@
...
+  inputs = {
+    mkWrapper.path = "/mkWrapper";
+    nixpkgs.path = "/nixpkgs";
+  };
+
+  options = {
+    homePath = {
+      type = types.pathLike;
+      description = ''
+        The path of `.irssi` to be injected into the wrapped package.
```
Squawkykaka: add a link to the documentation in  the style of

```
See `link to config` for valid options
```

modules/irssi.nix
```diff
@@ -0,0 +1,129 @@
...
+
+        Disjoint with the `configFile` option.
+      '';
+    };
+    configFile = {
+      type = types.pathLike;
+      description = ''
+        The config file to be injected into the wrapped package.
+
```
Squawkykaka, line 24: same as above

Squawkykaka: These also seem to do that same thing? both are paths

modules/irssi.nix
```diff
@@ -0,0 +1,129 @@
...
+        Disjoint with the `homePath` option.
+      '';
+    };
+
+    autoconnect = {
```
Squawkykaka, line 32: this can be done [without a union](https://github.com/adisbladis/korora?tab=readme-ov-file#features-1)

modules/irssi.nix
```diff
@@ -0,0 +1,129 @@
...
+  options = {
+    homePath = {
+      type = types.pathLike;
+      description = ''
+        The path of `.irssi` to be injected into the wrapped package.
+
+        Disjoint with the `configFile` option.
+      '';
+    };
+    configFile = {
```
Squawkykaka: There should preferabably be a `config` option that allows you to write in a string that then generates to the config, using a generator if its a specific format

Squawkykaka (changes requested):
Just some suggested changes, as it seems your missing the normal `config` options and only have the path ones.

modules/irssi.nix
poach3r: Doesn't this mean its possible for someone to define autoconnecting without a server and port?

modules/irssi.nix
poach3r: Irssi uses a strange format so I can't add a generator but I can allow strings.

modules/irssi.nix
llakala: Yeah, we'd want to avoid that. adding `types.either` would make this prettier, which is something I floated to adis before we fully forked.

modules/irssi.nix
```diff
@@ -0,0 +1,129 @@
...
+
+        See the documentation for valid options:                
+        https://irssi.org/documentation/settings/ 
+                                                                        
+        Disjoint with the `homePath` and `config` options.
+      '';
+    };
+
+    autoconnect = {
+      type = (
```
llakala: No need for parentheses here.

modules/irssi.nix
```diff
@@ -0,0 +1,129 @@
+{ types, ... }:
+{
+  inputs = {
+    mkWrapper.path = "/mkWrapper";
+    nixpkgs.path = "/nixpkgs";
+  };
+
+  options = {
+    homePath = {
```
llakala: I still don't understand why this is necessary. configFile allows using a homedir path thanks to types.pathLike. Maybe I'm missing something.

modules/irssi.nix
```diff
@@ -0,0 +1,129 @@
...
+      inherit (inputs.nixpkgs.pkgs) writeText;
+      inherit (inputs.nixpkgs.lib) optional optionals;
+      homeFlag = optional (options ? homePath) "--home=$out/irssi";
+      configFlag = optional (options ? configFile || options ? config) "config=$out/irssi/config";
+      autoconnectFlag = optionals (options ? autoconnect) (
+        [
+          "--connect=${options.autoconnect.server}"
+          "--port=${options.autoconnect.port}"
+        ]
+        ++ (optional (options.autoconnect ? password) "--password=${options.autoconnect.password}")
```
llakala: No parentheses needed around `optional` here.

modules/irssi.nix
llakala: Ah, I see. homePath lets us set all these files.
Image

Maybe we should just have `configDir` and `config` options. Then the user can use `configDir/config` if they want a config file. RFC42ing all the other files would be nice, but not necessary for this PR.

llakala (changes requested):

modules/irssi.nix
poach3r: Sounds good to me. I didn't think RFC42ing it was possible but after thinking about it more it should be possible if not tedious. I'll work on basic configuration now then RFC24 it in a future PR.

modules/irssi.nix
```diff
@@ -0,0 +1,112 @@
...
+  };
+
+  impl =
+    { options, inputs }:
+    let
+      inherit (builtins) concatStringsSep;
+      inherit (inputs.nixpkgs.pkgs) writeText;
+      inherit (inputs.nixpkgs.lib) optional optionals;
+      homeFlag = optional (options ? configDir) "--home=$out/irssi";
+      configFlag = optional (options ? config) "config=$out/irssi/config";
```
llakala: Seem to be missing a `--` here. I'll fix this myself and then merge.

modules/irssi.nix
```diff
@@ -0,0 +1,112 @@
...
+      preSymlink =
+        if options ? config then
+          ''
+            mkdir -p $out/irssi
+          ''
+        else
+          null;
+      symlinks = {
+        "$out/irssi/config" = if options ? config then writeText "config" options.config else null;
+        "$out/irssi" = if options ? homePath then options.homePath else null;
```
llakala: Also forgot to change this.

llakala (approved):
Would request changes, but I can fix these two things myself.

PR #41, opened 2026-03-22
nushell: fix regression in latest version

Squawkykaka:
## Checklist


Squawkykaka:
i ran nixfmt but it still fails

llakala:
> i ran nixfmt but it still fails

Maybe you're not in the devshell? I think it's the top of the file with `@`.

PR #44, opened 2026-04-22
nushell: rework to allow shellInit and path modification at the same …

Squawkykaka:
This rework updates the nushell module to use the source command in nushell to allow using the text based config editing and the sources at the same time

this still supports the toString method meaning it supports all the same stuff

## Checklist

- [ ] I checked [contributing.md](../docs/contributing.md) and my changes conform to it

### DIFF
--- a/modules/nushell.nix
+++ b/modules/nushell.nix
@@ -13,24 +13,27 @@
 
         See the nushell documentation for valid options:
         https://www.nushell.sh/book/configuration.html
-
-        Disjoint with the `configFile` option.
       '';
       mutatorType = types.string;
       mergeFunc = adios.lib.merge.strings.concatLines;
     };
-    configFile = {
-      type = types.pathLike;
+    sourceFiles = {
+      type = types.listOf types.pathLike;
       description = ''
-        `config.nu` file to be injected into the wrapped package.
-
-        See the nushell documentation on file syntax:
-        https://www.nushell.sh/book/configuration.html
-
-        Disjoint with the `config` option.
+        A list of paths to source in the `config.nu` file.
+        This can be used to import extra files, and can be used with impurity
       '';
+      mutatorType = types.listOf types.pathLike;
+      mergeFunc = adios.lib.merge.lists.concat;
     };
-
+    extraPackages = {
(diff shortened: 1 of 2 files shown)


modules/nushell.nix
```diff
@@ -44,13 +36,10 @@
...
 
-      configNu =
-        if options ? configFile then
-          options.configFile
-        else if options ? shellInit then
-          writeText "config.nu" options.shellInit
-        else
-          null;
+      assembledConfig = ''
+        ${builtins.concatStringsSep "\n" (map (el: "source ${el}") options.imports)}
```
llakala: Inherit builtins directly.

modules/nushell.nix
```diff
@@ -44,13 +36,10 @@
...
       inherit (inputs.nixpkgs.pkgs) writeText;
 
-      configNu =
-        if options ? configFile then
-          options.configFile
-        else if options ? shellInit then
-          writeText "config.nu" options.shellInit
-        else
-          null;
+      assembledConfig = ''
```
llakala: Doing this inside an indented string seems a little gross to me. How about something like:
```nix
configContents = concatStringsSep "\n" (
  map (file: "source ${file}") options.configFiles ++ (if options.shellInit then [ options.shellInit ] else [])
);
```
This prevents any extra newlines.

modules/nushell.nix
llakala: If it's not disjoint anymore, shouldn't this be removed? Be careful of things like this!

llakala (changes requested):
Left a response on discord proposing a slightly different change - these are just some issues I noticed that would need to be fixed anyways.

modules/nushell.nix
```diff
@@ -6,31 +6,23 @@
   };
 
   options = {
-    shellInit = {
+    imports = {
```
llakala: The word `imports` already has a context in the meaning of nix module systems. I'd prefer something like `configFiles`.

llakala (changes requested):
You accidentally committed the `.direnv` dir - I'll force push and fix this locally.

PR #51, opened 2026-05-22
modules/noctalia-shell: init

pengolord:
## Checklist

  - [x] settings variant
  - [ ] path variant

Noctalia is in a transition period to a new config format, and the one in nixpkgs uses an outdated format, so I added a warning to the default package spec. 
I marked this pr as a draft in case we want to wait for the newer format to reach nixpkgs, which might take a while.

### DIFF
--- a/modules/noctalia-shell.nix
+++ b/modules/noctalia-shell.nix
@@ -1,0 +1,65 @@
+{ types, ... }:
+{
+  inputs = {
+    mkWrapper.path = "/mkWrapper";
+    nixpkgs.path = "/nixpkgs";
+  };
+
+  options = {
+    settings = {
+      type = types.attrs;
+      description = ''
+        Settings to be injected into the wrapped package's `config.toml`.
+
+        See the noctalia docs for valid options:
+        https://docs.noctalia.dev/v5
+
+        Disjoint with the `configFile` option.
+      '';
+    };
+    configFile = {
+      type = types.pathLike;
+      description = ''
+        `config.toml` file to be injected into the wrapped package.
+
+        See the noctalia docs for valid options:
+        https://docs.noctalia.dev/v5
+
+        Disjoint with the `settings` option.
+      '';
+    };
+    package = {
+      type = types.derivation;
+      description = "The noctalia package to be wrapped.";
+      defaultFunc =
+        { inputs }:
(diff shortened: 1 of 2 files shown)


modules/noctalia-shell.nix
```diff
@@ -0,0 +1,65 @@
...
+
+        Disjoint with the `settings` option.
+      '';
+    };
+    package = {
+      type = types.derivation;
+      description = "The noctalia package to be wrapped.";
+      defaultFunc =
+        { inputs }:
+        inputs.nixpkgs.lib.warn "adios-wrappers: as of writing, noctalia-shell from nixpkgs uses noctalia v4, which this module is not compatible with. Please set `options.noctalia.package` to a package of noctalia v5 if this is still the case." inputs.nixpkgs.pkgs.noctalia-shell;
```
Squawkykaka: This could just be done with `builtins.warn`
```nix
builtins.warn "description" inputs.nixpkgs.pkgs.noctalia-shell
```

Squawkykaka (approved):
Looks good apart from 1 small nit

modules/noctalia-shell.nix
llakala: Yes, but `builtins.warn` was only added to a recent lix version - if someone's on an older version, they may not have it yet.

modules/noctalia-shell.nix
```diff
@@ -0,0 +1,65 @@
...
+      inherit (options) package;
+      preSymlink = ''
+        mkdir -p $out/noctalia
+      '';
+      symlinks = {
+        "$out/noctalia/noctalia.toml" =
+          if options ? configFile then
+            options.configFile
+          else if options ? settings then
+            generator.generate options.settings
```
llakala, line 57: This looks wrong to me - typically generators need a filename passed - like `generator.generate "yazi.toml" options.settings`.

llakala (changes requested):
I might want to wait until there's a package in nixpkgs that we can set as a default.

bivsk:
Seems like v5 has [made its way to unstable](https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/by-name/no/noctalia/package.nix)

Should probably reflect the upstream name change and rename this module to `noctalia`, but otherwise looks good. I have been using for a few weeks.

bivsk:
Hey @pengolord , are you still using this module? I'd be happy to maintain this if not, but hoping to get it merged soon either way.

PR #52, opened 2026-06-04
docs/nonflakes: Use main instead of master branch

Coca162:
Assuming it got renamed without this being updated

## Checklist

N/A


llakala:
Thanks!

PR #53, opened 2026-06-06
modules/fastfetch: init

Coca162:
## Checklist

### DIFF
--- a/modules/fastfetch.nix
+++ b/modules/fastfetch.nix
@@ -1,0 +1,67 @@
+{ types, ... }:
+{
+  inputs = {
+    mkWrapper.path = "/mkWrapper";
+    nixpkgs.path = "/nixpkgs";
+  };
+
+  options = {
+    settings = {
+      type = types.attrs;
+      description = ''
+        Settings to be injected into the wrapped package's `config.jsonc`.
+
+        See the documentation for valid options:
+        https://github.com/fastfetch-cli/fastfetch/wiki/Configuration
+
+        Disjoint with the `configFile` option.
+      '';
+    };
+    configFile = {
+      type = types.pathLike;
+      description = ''
+        `config.jsonc` file to be injected into the wrapped package.
+
+        See the documentation for syntax and valid options:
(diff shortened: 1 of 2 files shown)


modules/fastfetch.nix
```diff
@@ -0,0 +1,68 @@
...
+        "$out/fastfetch/config.jsonc" =
+          if options ? configFile then
+            options.configFile
+          else if options ? settings then
+            writeText "config.jsonc" (toJSON options.settings)
+          else
+            null;
+      };
+
+      flags = [
```
llakala: Nitpick - set this to only be defined if `options ? configFile || options ? settings`.

modules/fastfetch.nix
```diff
@@ -0,0 +1,68 @@
...
+  impl =
+    { options, inputs }:
+    let
+      inherit (inputs.nixpkgs.pkgs) writeText;
+      inherit (builtins) toJSON;
+    in
+    assert !(options ? settings && options ? configFile);
+    inputs.mkWrapper {
+      inherit (options) package;
+
```
llakala: I'd remove the newlines above and below the symlinks attribute.

modules/fastfetch.nix
```diff
@@ -0,0 +1,68 @@
...
+      symlinks = {
+        "$out/fastfetch/config.jsonc" =
+          if options ? configFile then
+            options.configFile
+          else if options ? settings then
+            writeText "config.jsonc" (toJSON options.settings)
```
llakala: We could use `pkgs.writers.writeJSON` here. I'll do some PRs to get us into using writers more.

llakala (changes requested):
Just small stuff.

modules/fastfetch.nix
llakala: Actually, let's use `pkgs.formats.json {}`. Formats seem to have the most options.

modules/fastfetch.nix
Coca162: I assume its fine to have this be a optionals instead of completely not setting flags

llakala:
Thanks!

PR #55, opened 2026-06-06
modules/btop: init

Coca162:
## Checklist

### DIFF
--- a/modules/btop.nix
+++ b/modules/btop.nix
@@ -1,0 +1,86 @@
+{ types, ... }:
+{
+  inputs = {
+    mkWrapper.path = "/mkWrapper";
+    nixpkgs.path = "/nixpkgs";
+  };
+
+  options = {
+    settings = {
+      type = types.attrs;
+      description = ''
+        Settings to be injected into the wrapped package's `btop.conf`.
+
+        See the documentation for valid options:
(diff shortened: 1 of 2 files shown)


modules/btop.nix
```diff
@@ -0,0 +1,90 @@
...
+      inherit (inputs.nixpkgs.lib) generators;
+
+      # Mostly copied from home-manager
+      # https://github.com/nix-community/home-manager/blob/master/modules/programs/mpv.nix
+      toKeyValue = generators.toKeyValue {
+        mkKeyValue = generators.mkKeyValueDefault {
+          mkValueString =
+            v:
+            let
+              inherit (builtins) isBool isString;
```
llakala: I'd inherit these builtins at the toplevel of the `impl` scope, otherwise they'll be redefined on every value.

modules/btop.nix
```diff
@@ -0,0 +1,90 @@
...
+              (if v then "True" else "False")
+            else if isString v then
+              ''"${v}"''
+            else
+              toString v;
+        } " = ";
+      };
+    in
+    assert !(options ? settings && options ? configFile);
+    inputs.mkWrapper {
```
llakala, line 66: See my comments on the other PRs.

llakala (changes requested):

PR #56, opened 2026-06-06
modules/mpv: init

Coca162:
## Checklist

### DIFF
--- a/modules/mpv.nix
+++ b/modules/mpv.nix
@@ -1,0 +1,125 @@
+{ types, ... }:
+{
+  inputs = {
+    mkWrapper.path = "/mkWrapper";
+    nixpkgs.path = "/nixpkgs";
+  };
+
+  options = {
+    settings = {
+      type = types.attrs;
+      description = ''
+        Settings to be injected into the wrapped package's `mpv.conf`.
+
+        See the manual for valid options:
+        https://mpv.io/manual/stable/
+
+        Disjoint with the `configFile` option.
+      '';
+    };
+    configFile = {
+      type = types.pathLike;
+      description = ''
+        `mpv.conf` file to be injected into the wrapped package.
+
+        See the manual for syntax and valid options:
+        https://mpv.io/manual/stable/
+
+        Disjoint with the `settings` option.
+      '';
+    };
+
+    keybinds = {
+      type = types.attrs;
+      description = ''
+        Keybinds to be injected into the wrapped package's `input.conf`.
+
+        See the manual for valid options:
+        https://mpv.io/manual/stable/#input-conf
+
+        Disjoint with the `keybindsFile` option.
+      '';
+    };
+    keybindsFile = {
+      type = types.pathLike;
+      description = ''
+        `input.conf` file to be injected into the wrapped package.
+
+        See the manual for syntax and valid options:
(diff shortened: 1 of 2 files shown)


modules/mpv.nix
```diff
@@ -0,0 +1,130 @@
...
+        `mpv.conf` file to be injected into the wrapped package.
+
+        See the manual for syntax and valid options:
+        https://mpv.io/manual/stable/
+
+        Disjoint with the `settings` option.
+      '';
+    };
+
+    bindings = {
```
llakala: Maybe name this `keybinds` and `keybindsFile`? I get why you named it `inputFIle`, but I think `bindings` is a little gross.

modules/mpv.nix
```diff
@@ -0,0 +1,130 @@
...
+      type = types.attrs;
+      description = ''
+        Key bindings to be injected into the wrapped package's `input.conf`.
+
+        See the manual for valid options:
+        https://mpv.io/manual/stable/#input-conf
+
+        Disjoint with the `configFile` option.
```
llakala: Copy/paste gone wrong.

modules/mpv.nix
```diff
@@ -0,0 +1,130 @@
...
+
+  impl =
+    { options, inputs }:
+    let
+      inherit (inputs.nixpkgs.pkgs) writeText;
+      inherit (inputs.nixpkgs.lib) generators concatStringsSep mapAttrsToList;
+      inherit (builtins) typeOf stringLength;
+
+      # Most of this copied from https://github.com/nix-community/home-manager/blob/master/modules/programs/mpv.nix
+
```
llakala: Nitpick - no newline here.

modules/mpv.nix
```diff
@@ -0,0 +1,130 @@
...
+      renderBindings =
+        bindings: concatStringsSep "\n" (mapAttrsToList (name: value: "${name} ${value}") bindings);
+    in
+    assert !(options ? settings && options ? configFile);
+    assert !(options ? bindings && options ? inputFile);
+    inputs.mkWrapper {
+      inherit (options) package;
+
+      name = "mpv";
```
llakala: Is the name necessary here? I'd also put it first, with less newlines inside the mkWrapper call.

llakala (changes requested):

modules/mpv.nix
Coca162: `pkgs.mpv` is already wrapped to provide scripts to it so if it were to be passed through without a name it'd error as the pname is `mpv-with-scripts`. Not sure how sinful it is to double wrap but hey

modules/mpv.nix
Coca162: I'm not that particular on the names, I did also think `inputFile` might've been a bit confusing with `inputs`

modules/mpv.nix
llakala: Sounds good. It might be slightly nicer to call mpv-unwrapped directly, but I'm not picky.

Coca162: I'll make it the default to use mpv-unwrapped but keep this for wrapped script users (at a later point maybe replicating the [script wrapping](https://discord.com/channels/976120247427948564/986671988720414820/1513033796688150578) in this wrapper instead)

PR #63, opened 2026-06-17
modules/niri: init

Coca162:
## Checklist

### DIFF
--- a/modules/niri.nix
+++ b/modules/niri.nix
@@ -1,0 +1,55 @@
+{ types, ... }: {
+  inputs = {
+    mkWrapper.from = { parent }: parent.mkWrapper;
+    nixpkgs.from = { parent }: parent.nixpkgs;
+  };
+
+  options = {
+    configFile = {
+      type = types.pathLike;
+      description = ''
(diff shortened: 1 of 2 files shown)


modules/niri.nix
```diff
@@ -0,0 +1,55 @@
+{ types, ... }: {
+  inputs = {
+    mkWrapper.from = { parent }: parent.mkWrapper;
+    nixpkgs.from = { parent }: parent.nixpkgs;
+  };
+
+  options = {
+    configFile = {
```
llakala, line 8: Thoughts on a `configLines` option? Maybe unnecessary, I just know hjem-rum does it.

llakala (approved):
Non-blocking, just curious.

modules/niri.nix
Coca162: Seems they append to some options they provide, feels kinda hacky. We could have a string option but if someone wanted to pass in the entire config through a nix string then I think slightly hardening it by requiring them to themselves turn it into a file isn't going to be too bad (especially for something that should be rare with how long these files tend to be)

llakala: Sounds good.

PR #64, opened 2026-06-18
modules/gitui: init

aabush64:
## Checklist

### DIFF
--- a/modules/gitui.nix
+++ b/modules/gitui.nix
@@ -1,0 +1,230 @@
+{ types, ... }: {
+  inputs = {
+    mkWrapper.from = { parent }: parent.mkWrapper;
+    nixpkgs.from = { parent }: parent.nixpkgs;
+  };
+
+  options = {
+    keybinds = {
+      type = types.attrsOf (
+        (types.struct "keybind" {
+          char = types.string;
+          mod = types.string;
+        }).override
+          { total = false; }
+      );
+      description = ''
+        Keybind overrides to be injected into the wrapped package's `key_bindings.ron`.
+
+        Each attrset in the optionAttrsets are converted to RON in this manner:
+        `''${action_name} = { char = "x"; mod = "y"; }; => action: Some(( code: x, modifiers: y))`
+
+        Where
+        `action_name` is a valid [Gitui action](https://github.com/gitui-org/gitui/blob/master/src/keys/key_list.rs#L83),
+        `char` is an *optional* valid [Crossterm KeyCode](https://docs.rs/crossterm/latest/crossterm/event/enum.KeyCode.html)
+        and `mod` is an *optional* valid [Crossterm Modifier](https://docs.rs/crossterm/latest/crossterm/event/struct.KeyModifiers.html)
+
+        An action that does not set either `char` or `mod` disables the keybind.
+
+        For more information on how `key_bindings.ron` is formatted and used:
+        https://github.com/gitui-org/gitui/blob/master/KEY_CONFIG.md
+
+        Disjoint with the `keybindsFile` option.
+      '';
+      example = {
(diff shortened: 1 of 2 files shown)


modules/gitui.nix
```diff
@@ -0,0 +1,193 @@
...
+          char = types.string;
+          mod = types.string;
+        }).override
+          { total = false; }
+      );
+      description = ''
+        Keybind overrides to be injected into the wrapped package's `key_bindings.ron`.
+
+        Each attrset in the optionAttrsets are converted to RON in this manner:
+        `action_name = { char = "x"; mod = "y"; }; => action: Some(( code: x, modifiers: y))`
```
llakala: Could we clarify that `char` and `mod` are optional, in a way that doesn't use the Some semantics?

modules/gitui.nix
```diff
@@ -0,0 +1,193 @@
...
+      '';
+    };
+
+    keySymbols = {
+      type = types.attrsOf types.string;
+      description = ''
+        Key symbol overrides to be injected into the wrapped package's `key_symbols.ron`.
+
+        Each attrset pair is converted to RON in this manner:
+        `{ key = symbol }; => key: Some(symbol)`
```
llakala: Rather than `key =`, maybe `${key} =`, to clarify that it's not the literal `key`?

modules/gitui.nix
```diff
@@ -0,0 +1,193 @@
...
+      inherit (options) package;
+      flags = if options ? flags then options.flags else [];
+      symlinks = {
+        "$out/gitui/key_bindings.ron" =
+          if options ? keyBindingsFile then
+            options.keyBindingsFile
+          else if options ? keyBindings then
+            writeText "key_bindings.ron" ''
+              (
+              ${keybindGenerator options.keyBindings}
```
llakala: if these are supposedly generators, maybe they should act more like them (not needing `writeText` to be called, and not needing this `()` logic? Alternatively, just rename them `makeKeybinds`, `makeThemes`, etc.

modules/gitui.nix
```diff
@@ -0,0 +1,193 @@
+{ types, ... }: {
+  inputs = {
+    mkWrapper.from = { parent }: parent.mkWrapper;
+    nixpkgs.from = { parent }: parent.nixpkgs;
+  };
+
+  options = {
```
llakala, line 7: Maybe use the `example` field for the `keyBindings`, `keySymbols`, and `theme` options?

llakala: Also, thoughts on the option names `keybinds` / `keybindsFile` and `symbols` / `symbolsFile`? Less like upstream, but matches other modules better.

llakala (changes requested):

modules/gitui.nix
aabush64: Works.

modules/gitui.nix
aabush64: Given their specificity probably both

modules/gitui.nix
aabush64: Can make that change, and yes they do need examples.

modules/gitui.nix
aabush64: Can clarify, actually brings up a need for extra things to fully implement Some()/None from ron

modules/gitui.nix
llakala: wdym both?

aabush64: The ( ) should be included in the let block but they aren't really "generators" so the names should be changed

llakala:
I added a commit with some cleanups. Let me know if these look good, and I can merge.

aabush64:
:+1:

PR #65, opened 2026-06-18
modules/wezterm: init

aabush64:
## Checklist

### DIFF
--- a/modules/wezterm.nix
+++ b/modules/wezterm.nix
@@ -1,0 +1,84 @@
+{ types, ... }: {
+  inputs = {
+    mkWrapper.from = { parent }: parent.mkWrapper;
+    nixpkgs.from = { parent }: parent.nixpkgs;
+  };
+
+  options = {
+    configContents = {
+      type = types.string;
+      description = ''
+        Lua configuration to be injected into the wrapped package's `wezterm.lua`.
+
+        See the documentation for valid options:
+        https://wezterm.org/config/lua/general.html
+
+        Disjoint with the `configFile` option.
+      '';
+    };
+
+    configFile = {
+      type = types.pathLike;
+      description = ''
+        `wezterm.lua` file to be injected into the wrapped package.
+
+        See the documentation for valid options:
+        https://wezterm.org/config/lua/general.html
+
+        Disjoint with the `configContents` option.
+      '';
+    };
+
+    configModules = {
+      type = types.listOf types.pathLike;
+      description = ''
+        Files containing .lua configuration modules to be injected into the wrapped package.
+
+        See the documentation to learn how to refer to modules in your wezterm.lua file:
+        https://wezterm.org/config/files.html?h=modules#making-your-own-lua-modules
+      '';
+    };
+
+    package = {
+      type = types.derivation;
+      description = "The wezterm package to be wrapped.";
+      defaultFunc = { inputs }: inputs.nixpkgs.pkgs.wezterm;
+    };
+  };
+
+  impl =
+    { options, inputs }:
+    let
+      inherit (builtins) listToAttrs;
+      inherit (inputs.nixpkgs.pkgs) writeText;
+    in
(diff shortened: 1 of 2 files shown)


modules/wezterm.nix
```diff
@@ -0,0 +1,122 @@
...
+      description = ''
+        Command line flag to set the used configuration to the pathLike provided by `configFile`.
+
+        Default behavior (false) uses the `WEZTERM_CONFIG_FILE` environment variable instead of the command line flag.
+
+        Requires that either `configContents` or `configFile` is set.
+      '';
+    };
+
+    skipConfig = {
```
llakala: Is this actually useful in a wrapper context?

modules/wezterm.nix
```diff
@@ -0,0 +1,122 @@
...
+      type = types.listOf types.pathLike;
+      description = ''
+        Files containing .lua configuration modules to be injected into the wrapped package.
+
+        See the documentation to learn how to refer to modules in your wezterm.lua file:
+        https://wezterm.org/config/files.html?h=modules#making-your-own-lua-modules
+      '';
+    };
+
+    configFileFlag = {
```
llakala: Same for this. Doesn't seem to be very useful.

modules/wezterm.nix
```diff
@@ -0,0 +1,122 @@
...
+          [
+            "--config-file"
+            "$out/wezterm-config/wezterm.lua"
+          ]
+        )
+        ++ (optionals (options ? skipConfig && options.skipConfig) [ "-c" ]);
+      environment = {
+        WEZTERM_CONFIG_FILE = "$out/wezterm-config/wezterm.lua";
+        # Lua is dum sometimes
+        LUA_PATH = "$out/wezterm-config/?.lua";
```
llakala, line 81: Does ? really work here?

modules/wezterm.nix
```diff
@@ -0,0 +1,122 @@
...
+            && options.configFileFlag
+          )
+          [
+            "--config-file"
+            "$out/wezterm-config/wezterm.lua"
+          ]
+        )
+        ++ (optionals (options ? skipConfig && options.skipConfig) [ "-c" ]);
+      environment = {
+        WEZTERM_CONFIG_FILE = "$out/wezterm-config/wezterm.lua";
```
llakala, line 79: Are we doing double-duty, using `--config-file` AND `WEZTERM_CONFIG_FILE`?

llakala (changes requested):

modules/wezterm.nix
aabush64: Its not, was just trying to cover bases.

modules/wezterm.nix
aabush64: See previous

modules/wezterm.nix
aabush64: We are, should probably delete

modules/wezterm.nix
llakala: I'd remove the two options, then.

modules/wezterm.nix
```diff
@@ -0,0 +1,95 @@
...
+      type = types.listOf types.pathLike;
+      description = ''
+        Files containing .lua configuration modules to be injected into the wrapped package.
+
+        See the documentation to learn how to refer to modules in your wezterm.lua file:
+        https://wezterm.org/config/files.html?h=modules#making-your-own-lua-modules
+      '';
+    };
+
+    flags = {
```
llakala: I think we can remove this too.

modules/wezterm.nix
```diff
@@ -0,0 +1,95 @@
...
+      type = types.listOf types.string;
+      description = ''
+        Flags to be appended to the wrapped package
```
llakala: Needs a period. Would fix myself if it was only this, but might as well.

llakala (changes requested):
Sorry to not catch this on first review!

PR #66, opened 2026-06-19
modules/dunst: init

aabush64:
## Checklist

SGMMSWMDLIMM

### DIFF
--- a/modules/dunst.nix
+++ b/modules/dunst.nix
@@ -1,0 +1,81 @@
+{ types, ... }: {
+  inputs = {
+    mkWrapper.from = { parent }: parent.mkWrapper;
+    nixpkgs.from = { parent }: parent.nixpkgs;
+  };
+
+  options = {
+    configContents = {
+      type = types.string;
+      description = ''
+        Settings to be injected into the wrapped package's `dunstrc`.
+
+        See the documentation for syntax and valid options:
+        https://dunst-project.org/documentation/
+
+        Disjoint with the `configFile` option.
+      '';
+    };
+    configFile = {
+      type = types.pathLike;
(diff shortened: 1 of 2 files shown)


aabush64:
Gonna have to start putting a nickel in a jar every time I have to force push fr

modules/dunst.nix
```diff
@@ -0,0 +1,92 @@
...
+      type = types.listOf types.pathLike;
+      description = ''
+        `*.conf` files to be injected into the wrapped package alongside the main configuration.
+
+        See the documentation for syntax and valid options:
+        https://dunst-project.org/documentation/
+      '';
+    };
+
+    flags = {
```
llakala: Don't think we need this. I prefer specialized options for flags when the program isn't entirely flag-based in its config.

modules/dunst.nix
```diff
@@ -0,0 +1,92 @@
...
+      '';
+    };
+
+    dropinFiles = {
+      type = types.listOf types.pathLike;
+      description = ''
+        `*.conf` files to be injected into the wrapped package alongside the main configuration.
+
+        See the documentation for syntax and valid options:
+        https://dunst-project.org/documentation/
```
llakala: I think we can do better with this doc link. https://dunst-project.org/documentation/dunst/#FILES seems much more relevant.

llakala (changes requested):

llakala:
Better descriptions on the flag options would be nice. But to be clear, I don't actually think we need these, if they're not a common part of configuring dunst. I was referring to my preference in general, and I think all of those flags are niche enough that the user can pass them if necessary.

PR #67, opened 2026-06-21
modules/hydrus: init

aabush64:
## Checklist

### DIFF
--- a/modules/hydrus.nix
+++ b/modules/hydrus.nix
@@ -1,0 +1,60 @@
+{ types, ... }:
+{
+  inputs = {
+    mkWrapper.from = { parent }: parent.mkWrapper;
+    nixpkgs.from = { parent }: parent.nixpkgs;
+  };
+
+  options = {
+    clientFlags = {
+      type = types.listOf types.string;
+      description = ''
+        List of hydrus-client flags to pass to hydrus client on startup.
+
+        Note that paths passed to db_dir/temp_dir should be writable (not in the nix store)
+
+        See the documentation for valid syntax and formatting of the related flags:
+        https://hydrusnetwork.github.io/hydrus/launch_arguments.html
+      '';
+    };
+    serverFlags = {
+      type = types.listOf types.string;
+      description = ''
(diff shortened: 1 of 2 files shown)


modules/hydrus.nix
```diff
@@ -0,0 +1,151 @@
+{ types, ... }: {
+  inputs = {
+    mkWrapper.from = { parent }: parent.mkWrapper;
+    nixpkgs.from = { parent }: parent.nixpkgs;
+  };
+
+  options = {
+    clientFlags = {
```
llakala, line 9: If we're going to name the option `clientFlags` and `serverFlags`, maybe it should just be a `listOf string`. We could use the current form with `clientSettings` and `serverSettings`, but I don't like the struct approach here, since it forces us to constantly provide a value for all the options they provide. If we were to use the attrset approach, then I'd prefer we use something like `lib.cli.toCommandLineGNU`. This would allow the type to be `attrs`, and would automatically turn attrs into the relevant commandline format. But we don't use that in any other modules, so for now let's stick to `listOf string`.

llakala (changes requested):

modules/hydrus.nix
aabush64: "forces us to constantly provide a value for all options they provide", doesn't `total = false;` in the struct negate that?

llakala: But we still need to add the logic in the `impl`.

llakala: With `lib.cli.toCommandLine`, we could just do:
```nix
{
  some-bool-field = true;
  some-string-field = "hi";
}
```
and it automatically turns the input into the relevant flag representations.

aabush64: Ah I see, then I agree with the last line of the first comment. I'd prefer to just rewrite it to listOf strings because intermediate guardrailing feels like it introduces more ambiguity than helps.

llakala: Yeah, until we standardize on using a function like this, listOf string.

llakala (approved):
Gonna fix some whitespace and merge.

PR #70, opened 2026-06-23
modules/zathura: init

bivsk:
## Checklist

Not 100% sure I'm doing doing the impure variant correctly, but please lmk! 

Also, I used `mappings` due to that being the term the zathura docs and config use. Open to changing to `keybinds` or something similar.

### DIFF
--- a/modules/zathura.nix
+++ b/modules/zathura.nix
@@ -1,0 +1,118 @@
+{ types, ... } @ adios:
+{
+  inputs = {
+    mkWrapper.from = { parent }: parent.mkWrapper;
+    nixpkgs.from = { parent }: parent.nixpkgs;
+  };
+
+  options = {
+    settings = {
+      type = types.attrs;
+      description = ''
+        Options to be injected into the wrapped package's `zathurarc`.
+
+        See the zathurarc documentation for the full list of options:
+        {manpage}`zathurarc(5)`.
+
+        Disjoint with the `sourceFiles` option.
+      '';
+      example = {
+        default-bg = "#000000";
+        default-fg = "#FFFFFF";
+      };
+      mutatorType = types.attrs;
+      mergeFunc = adios.lib.merge.attrs.recursively;
+    };
+    keybinds = {
+      type = types.attrs;
+      description = ''
+        Keybindings to be injected into the wrapped package's `zathurarc`.
+
+        See the zathurarc documentation for the full list of possible mappings:
+        {manpage}`zathurarc(5)`.
+
+        You can create a mode-specific mapping by specifying the mode before the key:
+        `"[normal] <C-b>" = "scroll left";`.
+
+        Disjoint with the `sourceFiles` option.
+      '';
+      example = {
+        D = "toggle_page_mode";
+        "<Right>" = "navigate next";
+        "[fullscreen] <C-i>" = "zoom in";
+      };
+      mutatorType = types.attrs;
+      mergeFunc = adios.lib.merge.attrs.recursively;
+    };
+    sourceFiles = {
+      type = types.listOf types.pathLike;
+      description = ''
+        A list of paths to source in the `zathurarc` file.
+        This can be used to import extra files, and can be used with impurity.
+
+        See the zathurarc documentation for valid options:
+        {manpage}`zathurarc(5)`.
+
+        Disjoint with `settings` and `keybinds` options.
+      '';
+      mutatorType = types.listOf types.pathLike;
+      mergeFunc = adios.lib.merge.lists.concat;
+    };
+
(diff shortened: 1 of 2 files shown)


modules/zathura.nix
```diff
@@ -0,0 +1,117 @@
+{ types, ... } @ adios:
+{
+  inputs = {
+    mkWrapper.from = { parent }: parent.mkWrapper;
+    nixpkgs.from = { parent }: parent.nixpkgs;
+  };
+
+  options = {
+    configLines = {
```
llakala: This should be named `settings`, not `configLines`.

modules/zathura.nix
```diff
@@ -0,0 +1,117 @@
...
+        Disjoint with the `sourceFiles` option.
+      '';
+      example = {
+        default-bg = "#000000";
+        default-fg = "#FFFFFF";
+      };
+      mutatorType = types.attrs;
+      mergeFunc = adios.lib.merge.attrs.recursively;
+    };
+    mappings = {
```
llakala: I get why you'd name it `mappings`, but I think consistency with other modules is preferred, so I'd rename it to `keybinds` and `keybindsFile`.

modules/zathura.nix
```diff
@@ -0,0 +1,117 @@
...
+        n: v:
+        let
+          formatValue = v: if isBool v then boolToString v else toString v;
+        in
+        ''set ${n}	"${formatValue v}"'';
+
+      formatMapLine = n: v: "map ${n}   ${toString v}";
+
+      assembledConfig = concatStringsSep "\n" (
+        optionals (options ? sourceFiles) (map (el: "include ${el}") options.sourceFiles)
```
llakala: Nitpick - since `sourceLines` is disjoint from the others, i'd do an if/else on whether `options ? sourceLines`. Otherwise we're always doing three concatenations, when the maximum is either one or two.

modules/zathura.nix
```diff
@@ -0,0 +1,117 @@
...
+        ++ optionals (options ? configLines) (mapAttrsToList formatLine options.configLines)
+        ++ optionals (options ? mappings) (mapAttrsToList formatMapLine options.mappings)
+      );
+    in
+    assert !(options ? sourceFiles && (options ? configLines || options ? mappings));
+    inputs.mkWrapper {
+      name = "zathura";
```
llakala, line 96: Is the package not already named zathura? This option is only needed in certain cases where the pname doesn't match.

modules/zathura.nix
```diff
@@ -0,0 +1,117 @@
...
+      inherit (options) package;
+      symlinks = {
+        "$out/zathura/zathurarc" = writeText "zathurarc" assembledConfig;
```
llakala, line 99: Technically, we could check if _any_ of the options are set, and not write the file if not. But I won't block this on that.

llakala (changes requested):
Looks good other than the few notes.

modules/zathura.nix
bivsk: Sounds good. Should I create a separate `keybindsFile`? I assumed users can use `sourceFiles` for both files that contain keymaps and those with settings.

llakala: Oh, oops, good point. Yeah, no need for a separate option, then.

modules/zathura.nix
bivsk: I believe most users use the package `zathura-with-plugins`

llakala: Then looks good.

modules/zathura.nix
bivsk: I think it _may_ make more sense to leave this as is, since even in cases where we're using the impure `sourceFiles`, I am importing them via `include ${file}` lines within the `zathurarc`.

In a case where literally no options are set, is the wrapper any different from the default `zathura` package? No problem making the change, I just don't quite understand why or when a user would not set any options (impure or not).

modules/zathura.nix
bivsk: Let me know if I understood correctly here.

modules/zathura.nix
llakala: > In a case where literally no options are set, is the wrapper any different from the default zathura package? No problem making the change, I just don't quite understand why or when a user would not set any options (impure or not).

Yeah, it's not. Sometimes I try to work around this in the modules, but we should really just have better assertion logic, and state an XOR of the disjoint options, rather than just a NAND. Not actually blocking.

llakala:
LGTM. I'm gonna force push to fix some style nits, then merge.

llakala:
Thanks for the module!

PR #71, opened 2026-06-24
modules/eza: Fix option naming and environment variables

aabush64:
## Checklist


llakala (approved):

llakala:
We should probably add adios logic for deprecating an option at some point, rather than just removing them.

PR #72, opened 2026-06-25
docs: update

aabush64:
Adding more styling guidelines to `contributing.md` and fixing mkWrapper option docs.


aabush64:
Ooops that was not supposed to be pushed here

PR #75, opened 2026-06-27
rework _get-docs.nix to allow using ndg in future

Squawkykaka:
## Checklist

- [ ] I checked [contributing.md](../docs/contributing.md) and my changes conform to it


llakala:
We should probably get rid of some (if not all) of the sed replacements in generate-docs.sh, if we're no longer going to be using it as human-readable text. This should also preserve newlines in the generated descriptions, which I care more about.

Squawkykaka:
Right i think this is ready to be merged now, and also allows futhur work on ndg and documentation to continue.

Ive decided to remove options.json from the repo, as it is no longer a reference and instead should be generated whenever things change #78 has furthur work on this

llakala:
Im not sure on merging this, as my experience with ndg has been rather negative so far, and I dislike the lack of pagination per module. I was planning on looking into a basic SSG a la Astro.

llakala:
Closing this in favor of #79.

PR #77, opened 2026-06-29
modules/ripgrep-all: init

aabush64:
## Checklist

### DIFF
--- a/modules/ripgrep-all.nix
+++ b/modules/ripgrep-all.nix
@@ -1,0 +1,117 @@
+{ types, ... } @ adios:
+{
+  inputs = {
+    mkWrapper.from = { parent }: parent.mkWrapper;
+    nixpkgs.from = { parent }: parent.nixpkgs;
+  };
+
+  options = {
+    flags = {
+      type = types.listOf types.string;
+      description = ''
+        Flags to be automatically appended when running ripgrep-all.
+
+        See the documentation of valid flags:
+        https://github.com/phiresky/ripgrep-all#flags
+
+        Disjoint with the `configFile` option.
+      '';
+    };
+
+    adapters = {
+      type = types.listOf types.attrs;
+      description = ''
+        Custom adapters to be appended to the wrapped package's `config.jsonc` after the `configContents` option.
+
+        See the documentation for valid options:
+        https://github.com/phiresky/ripgrep-all/wiki#custom-adapters
+
+        Disjoint with the `configFile` option.
+      '';
+      mutatorType = types.listOf types.attrs;
+      mergeFunc = adios.lib.merge.lists.concat;
+      example = [
+        {
+          name = "poppler";
+          version = 1;
+          description = "Uses pdftotext (from poppler-utils) to extract plain text from PDF files";
+
+          extensions = [ "pdf" ];
+          mimetypes = [ "application/pdf" ];
+
+          binary = "pdftotext";
+          args = [
+            "-"
+            "-"
+          ];
+          disabled_by_default = false;
+          output_path_hint = "\${input_virtual_path}.txt.asciipagebreaks";
+        }
+      ];
+    };
+
+    configFile = {
+      type = types.pathLike;
+      description = ''
+        `config.jsonc` file to be injected into the wrapped package.
+
+        See the documentation for syntax and valid options:
+        https://github.com/phiresky/ripgrep-all#flags
+
+        The json schema can be also be accessed via `rga --rga-print-json-schema` for more formatting detail.
+
+        Disjoint with the `flags` and `adapters` option.
+      '';
+    };
+
+    package = {
+      type = types.derivation;
+      description = ''
+        The `ripgrep-all` package to be wrapped.
+
+        As `ripgrep-all` makes use of the `ripgrep` package (also available in adios-wrappers), you can change the ripgrep package to your own via an injection.
+
+        Overrides can be accomplished in adios-wrappers in the following manner:
+
+        1. Add the `ripgrep` wrapper to the module inputs:
+        `inputs.ripgrep.from = { parent }: parent.ripgrep;`
+
+        2. Set the package option to an overriden `ripgrep-all` package with the previously used input:
+        `options.package.defaultFunc = { inputs }: inputs.nixpkgs.ripgrep-all.override { ripgrep = inputs.ripgrep; };`
+      '';
+      defaultFunc = { inputs }: inputs.nixpkgs.pkgs.ripgrep-all;
+    };
+  };
+
+  impl =
+    { options, inputs }:
+    let
+      inherit (builtins) toJSON;
+      inherit (inputs.nixpkgs.pkgs) formats;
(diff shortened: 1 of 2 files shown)


modules/ripgrep-all.nix
```diff
@@ -0,0 +1,146 @@
...
+        https://github.com/phiresky/ripgrep-all#flags
+
+        The json schema can be also be accessed via `rga --rga-print-json-schema` for more formatting detail.
+      '';
+    };
+
+    package = {
+      type = types.derivation;
+      description = "The ripgrep-all package to be wrapped.";
+      defaultFunc = { inputs }: inputs.nixpkgs.pkgs.ripgrep-all.override;
```
llakala: Well, no, we can't have the default be a link to a `.override`.

modules/ripgrep-all.nix
```diff
@@ -0,0 +1,146 @@
...
+  };
+
+  impl =
+    { options, inputs }:
+    let
+      inherit (builtins) toJSON;
+      inherit (inputs.nixpkgs.pkgs) writeText;
+      inherit (inputs.nixpkgs.lib) optionals;
+      flags = optionals (options ? flags) options.flags;
+      configContents = writeText "config.jsonc" (
```
llakala: Use `pkgs.formats.json`.

modules/ripgrep-all.nix
llakala: Also, could we add a note here about how to override to use the rg package?

modules/ripgrep-all.nix
```diff
@@ -0,0 +1,146 @@
...
+
+        The json schema can be also be accessed via `rga --rga-print-json-schema` for more formatting detail.
+
+        Custom adapters in `config.jsonc` are set via the `configAdapters` options.
+
+        Disjoint with the `flags` and `configFile` option.
+      '';
+    };
+    configAdapters =
+      let
```
llakala: I would honestly get rid of this option, since I don't trust the struct to remain current, and the custom adapters can already be set via the configFile.

modules/ripgrep-all.nix
```diff
@@ -0,0 +1,146 @@
...
+        Flags to be automatically appended when running ripgrep-all.
+
+        See the documentation of valid flags:
+        https://github.com/phiresky/ripgrep-all#flags
+
+        Disjoint with the `configContents` and `configFile` option.
+      '';
+    };
+
+    configContents = {
```
llakala: Is there anything that can be done by this option that `flags` can't already do? If not, I would just remove this.

modules/ripgrep-all.nix
```diff
@@ -0,0 +1,146 @@
...
+            options.configFile
+          else
+            null;
+      };
+
+      environment = {
+        XDG_CONFIG_HOME = "$out";
+      };
+    };
+}
```
llakala, line 117: Needs meta.maintainers.

llakala (changes requested):

modules/ripgrep-all.nix
aabush64: Override unintentional, on the second we could but technically any wrapper could have their package overriden its not a rga specific thing.

modules/ripgrep-all.nix
aabush64: Flags can't create adapters only the config file, if the adapters are gone then it leaves just this to create them.

modules/ripgrep-all.nix
aabush64: Why the lack of trust? This package does not get updated often, breaking changes less so.

modules/ripgrep-all.nix
llakala: Yeah, but we have an rg package in-tree, so I think it's worth noting for this specifically.

modules/ripgrep-all.nix
llakala: More of a policy I have treewide - I want to avoid locking us to verification structs that can go stale.

modules/ripgrep-all.nix
llakala: Oh, I see. Hmm, how about we keep `flags` and `configAdapters` as options that can both be set, and then have configFile as disjoint from both of them?

llakala: And maybe just name it `adapters`. I know, I know, I'm breaking the standard, but I like it.

modules/ripgrep-all.nix
```diff
@@ -0,0 +1,119 @@
...
+    in
+    assert !(options ? configFile && (options ? flags || options ? adapters));
+    inputs.mkWrapper {
+      inherit (options) package;
+      inherit flags;
+      name = "rga";
+      symlinks = {
+        "$out/ripgrep-all/config.jsonc" =
+          if options ? adapters then
+            writeText "config.jsonc" (toJSON {
```
llakala: Use `pkgs.formats`.

modules/ripgrep-all.nix
```diff
@@ -0,0 +1,119 @@
...
+    };
+  };
+
+  impl =
+    { options, inputs }:
+    let
+      inherit (builtins) toJSON;
+      inherit (inputs.nixpkgs.pkgs) writeText;
+      inherit (inputs.nixpkgs.lib) optionals;
+      flags = optionals (options ? flags) options.flags;
```
llakala: I'd just inline this.

modules/ripgrep-all.nix
```diff
@@ -0,0 +1,119 @@
...
+    in
+    assert !(options ? configFile && (options ? flags || options ? adapters));
+    inputs.mkWrapper {
+      inherit (options) package;
+      inherit flags;
+      name = "rga";
```
llakala: We no longer have a `name` option, thanks to coca. The option was renamed to `pname`, and it only changes the package name. You want `binaryName` - but I don't think it's necessary, because we now autodetect that based on `meta.mainProgram`.

modules/ripgrep-all.nix
```diff
@@ -0,0 +1,119 @@
...
+          if options ? adapters then
+            writeText "config.jsonc" (toJSON {
+              custom_adapters = options.adapters;
+            })
+          else if options ? configFile then
+            options.configFile
+          else
+            null;
+      };
+
```
llakala: No newline here.

llakala (changes requested):
Small stuff.

modules/ripgrep-all.nix
```diff
@@ -0,0 +1,117 @@
...
+        `options.package.defaultFunc = { inputs }: inputs.nixpkgs.ripgrep-all.override { ripgrep = inputs.ripgrep; };`
+      '';
+      defaultFunc = { inputs }: inputs.nixpkgs.pkgs.ripgrep-all;
+    };
+  };
+
+  impl =
+    { options, inputs }:
+    let
+      inherit (builtins) toJSON;
```
llakala, line 89: Forgot to remove this inherit.

PR #78, opened 2026-06-30
Setup ndg config file, and add some wiring for it

Squawkykaka:
This relies on the previous #75 for the options.json rework.

This pr sets up the ndg.toml file, and will also setup a github action to rebuild the documentation whenever options or docs change.

PR #79, opened 2026-07-01
site: init

llakala:
Still needs a lot of work - just a proof of concept. Alternative to #75 and #78.

## Checklist


Squawkykaka:
With this site, is it possible that a field be added for mutators to specify a description of what a mutator is doing? 

this could be done via something simlar to nixpkgs with something parsing and reading the documentation but due to adios modules having description fields in many situations it may make more sense to add a field to the mutators.

llakala:
Only issue would be where to store it. Right now, the schema is `mutations.$module.$option = function;`. We would need to change it to be like `mutations.$module.$option.computedValue = function;`.

PR #80, opened 2026-07-02
modules/mpv: wrap `umpv`

Coca162:
This unfortunately does not work when used in combination with the `mpv-with-scripts` nixpkgs wrapper but we don't default to it so whatever. At some point just copying over the entire thing into this wrapper might be the only nice route for users of it (me).

## Checklist

### DIFF
--- a/modules/mpv.nix
+++ b/modules/mpv.nix
@@ -113,6 +113,9 @@
           else
             null;
       };
+      postWrap = ''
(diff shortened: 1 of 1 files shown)


llakala:
Won't this conflict with the normal wrap, since you send it to $out/bin/mpv?

Coca162:
This should only set a environment variable to `$out/bin/mpv` and not send it to there, as far as I understand it.

Coca162:
Closed in favor of #81

PR #81, opened 2026-07-03
modules/mpv: Wrap with nixpkgs wrapper

Coca162:
After looking at it further I think this makes the most sense over having our own wrapper. 

## Checklist

### DIFF
--- a/modules/mpv.nix
+++ b/modules/mpv.nix
@@ -1,11 +1,18 @@
 { types, ... }:
 {
   inputs = {
-    mkWrapper.from = { parent }: parent.mkWrapper;
     nixpkgs.from = { parent }: parent.nixpkgs;
   };
 
   options = {
+    scripts = {
+      type = types.listOf types.derivation;
+      description = ''
+        Scripts to be added to the wrapped package.
+        Packaged scripts from nixpkgs can be found in `mpvScripts`.
+      '';
+    };
+
     settings = {
       type = types.attrs;
       description = ''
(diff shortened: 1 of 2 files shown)


modules/mpv.nix
```diff
@@ -92,30 +102,35 @@
...
-      };
+    mpv.override {
+      mpv-unwrapped = options.package;
+      scripts = options.scripts or [];
+      extraMakeWrapperArgs = [
+        "--add-flags"
+        (
+          "--no-config"
+          + (optionalString (configFile != null) " --include=${configFile}")
+          + (optionalString (keybindsFile != null) " --input-conf=${keybindsFile}")
```
llakala: Nit:
```suggestion
          + optionalString (configFile != null) " --include=${configFile}"
          + optionalString (keybindsFile != null) " --input-conf=${keybindsFile}"
```

modules/mpv.nix
```diff
@@ -92,30 +102,35 @@
...
-            options.keybindsFile
-          else if options ? keybinds then
-            writeText "input.conf" (renderKeybinds options.keybinds)
-          else
-            null;
-      };
-      environment = {
-        XDG_CONFIG_HOME = "$out";
-      };
+    mpv.override {
```
llakala, line 127: Comment explaining why we don't use `mkWrapper` here would be nice.

llakala (changes requested):
Looks good in concept.

llakala (approved):

PR #86, opened 2026-07-05
Futher work on astro site

Squawkykaka:
This has futhur work on #79, here so that people can check out improvements.

## Checklist
- [ ] add theming
- [ ] add some way to search modules, and options
- [ ] improve rendering of mutators
- [ ] render the description instead of leaving as raw text
  - could? be done as using ndg-commenmark, or more likely wiring into the render function in astro


llakala:
Accidentally committed .direnv.

Squawkykaka:
Ok so I think this is in a good place for a full review, I won't be handing styling much in this pr I'll work more on that later,

I'll consider this pr ready to merge with a GitHub action to build the doc site setup though I can do that separately if you perfer

llakala:
# Style nits
1. Could we decrease the gap after the toplevel heading, and generally after headers like Type? image

2. Can we also decrease the gap before the description term list? We may be accidentally preserving the final newline, unsure.
image

3. Less gap after Available Modules, and more gap between the modules? image

4. Let's just have no favicon for now instead of the Astro one.


# Non-style notes
1. Like I mentioned on Discord, I think we should just stop rendering the `Mutations` field for now. It's barely used, and there's not much good usecase for rendering it without a description.
2. Once we merge into my PR, I can fix the Nix rendering, so it doesn't just show as JSON. 
3. On your note, I think a separate PR for GHA (Github Actions) makes sense. I'd like to merge this in to my PR once my comments here are addressed, do some Nix-end cleanups on my end, and then merge into main. GHA PR can target main after that.

Squawkykaka:
@llakala this has the changes you mentioned, and should be fine to merge into your branch now.

PR #91, opened 2026-07-18
quickshell: init wrapper

Squawkykaka:
Add a wrapper for quickshell.
ive chosen not to add a nix-based builder for this, as it doesnt make sense for quickshell.

## Checklist

### DIFF
--- a/modules/quickshell.nix
+++ b/modules/quickshell.nix
@@ -42,6 +42,6 @@
     };
 
   meta = {
-    maintainers = [ "squawky" ];
+    maintainers = [ "Squawkykaka" ];
   };
 }
--- a/modules/quickshell.nix
+++ b/modules/quickshell.nix
@@ -1,0 +1,47 @@
+{ types, ... }:
+{
+  inputs = {
+    mkWrapper.from = { parent }: parent.mkWrapper;
+    nixpkgs.from = { parent }: parent.nixpkgs;
+  };
+
+  options = {
+    configDir = {
+      type = types.pathLike;
+      description = ''
+        Folder containing quickshell configuration files to be injected into the wrapped package.
+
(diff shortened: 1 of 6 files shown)


modules/quickshell.nix
```diff
@@ -0,0 +1,40 @@
...
+    package = {
+      type = types.derivation;
+      defaultFunc = { inputs }: inputs.nixpkgs.pkgs.quickshell;
+      description = "The quickshell package to be wrapped.";
+    };
+  };
+
+  impl =
+    { options, inputs }:
+    if options ? configDir then
```
llakala: Let's assert rather than being conditional. I have an idea to swap all our treewide instances of
```nix
  assert !(options ? settings && options ? configFile);
```
for:
```nix
  assert (options ? settings) != (options ? configFile);
```
XOR rather than NAND will let us assume that at least one is set.

modules/quickshell.nix
```diff
@@ -0,0 +1,40 @@
...
+      inputs.mkWrapper {
+        inherit (options) package;
+        symlinks = {
+          "$out/quickshell" = options.configDir;
+        };
+        environment = {
+          XDG_CONFIG_HOME = "$out";
```
llakala: Is there a flag we could use instead?

modules/quickshell.nix
```diff
@@ -0,0 +1,40 @@
...
+        };
+      }
+    else
+      options.package;
+}
```
llakala, line 47: Add `meta.maintainers`.

llakala (changes requested):

modules/quickshell.nix
Squawkykaka: I decided to use `asset options ? configDir` since there is only 1 option.

PR #94, opened 2026-07-29
modules/niri: change configFile to configFiles

bivsk:
My goal is to have the ability to declaratively manage my niri config, and allow mutations from other modules.  KDL does not represent well in a `settings` nix option, so I wanted to do something like `configContents`. However, while KDL itself allows for duplicate sections, niri performs its own checks and only allows certain sections to be duplicated. The exception to this rule is when the duplicated sections are split into several files.

This change creates a `sourceFiles` options, which takes a list of paths. These paths are then imported into the entrypoint config.kdl file. I have used `optional=true` to avoid warnings when impure paths are non-existent.  Modifiers of this option would need to utilize `writeText`, rather than simply providing a string of kdl lines.

I was considering modifying the existing `configFile` option, but went this route to avoid breaking anyone's config, however I am open to changing that option's type to a union of `types.pathLike` | `types.listOf types.pathLike`. Either route makes sense to me - open to feedback.

If something more robust, like [here](https://github.com/Lassulus/wrappers/blob/f57bc81aa9669871343341b13aa175c8927b6a74/modules/niri/module.nix#L8) is preferred to this route, please let me know.  This just seemed like the most straightforward and simplest way to provide a means of declaratively providing a niri config. 

## Checklist

### DIFF
--- a/modules/niri.nix
+++ b/modules/niri.nix
@@ -1,4 +1,4 @@
-{ types, ... }:
+{ types, ... } @ adios:
 {
   inputs = {
     mkWrapper.from = { parent }: parent.mkWrapper;
@@ -6,14 +6,15 @@
   };
 
   options = {
-    configFile = {
-      type = types.pathLike;
+    configFiles = {
+      type = types.listOf types.pathLike;
       description = ''
-        `config.kdl` file to be injected into the wrapped package.
+        A list of `.kdl` files that are included in the wrapped package's config.
 
         See the niri documentation for syntax and valid options:
         https://github.com/niri-wm/niri/wiki/Configuration:-Introduction
       '';
+      mergeFunc = adios.lib.merge.lists.concat;
     };
 
     package = {
@@ -25,7 +26,19 @@
 
   impl =
     { options, inputs }:
-    if options ? configFile then
+    if options ? configFiles then
+      let
+        inherit (builtins) concatStringsSep head length;
+        inherit (inputs.nixpkgs.pkgs) writeText;
+
+        configPath =
+          if length options.configFiles == 1 then
+            head options.configFiles
+          else
+            writeText "config.kdl" (
+              concatStringsSep "\n" (map (file: ''include "${file}"'') options.configFiles)
+            );
+      in
       inputs.mkWrapper {
         inherit (options) package;
         # Hack modified and gotten from https://github.com/Lassulus/wrappers/blob/main/modules/niri/module.nix
@@ -36,15 +49,15 @@
           [Service]
           ExecStart=
           ExecStart=$out/bin/niri --session
-          ExecReload=$out/bin/niri msg action load-config-file --path ${options.configFile}
+          ExecReload=$out/bin/niri msg action load-config-file --path ${configPath}
           X-ReloadIfChanged=true
           [Unit]
-          X-Reload-Triggers=${options.configFile}
+          X-Reload-Triggers=${configPath}
           EOF
           cp --remove-destination niri.service $out/share/systemd/user/niri.service
         '';
         environment = {
-          NIRI_CONFIG = options.configFile;
+          NIRI_CONFIG = configPath;
         };
       }
     else
--- a/docs/options.json
+++ b/docs/options.json
@@ -665,9 +665,9 @@
   },
 
   "niri": {
-    "configFile": {
-      "description": "`config.kdl` file to be injected into the wrapped package. See the niri documentation for syntax and valid options: https://github.com/niri-wm/niri/wiki/Configuration:-Introduction",
-      "type": "pathLike"
+    "configFiles": {
+      "description": "A list of `.kdl` files that are included in the wrapped package's config. See the niri documentation for syntax and valid options: https://github.com/niri-wm/niri/wiki/Configuration:-Introduction",
+      "type": "listOf<pathLike>"
     },
     "package": {
       "description": "The niri package to be wrapped.",


modules/niri.nix
```diff
@@ -16,6 +16,23 @@
...
 
+    sourceFiles = {
+      type = types.listOf types.pathLike;
+      description = ''
+        A list of kdl files to import into the wrapped package's config.
+
+        Niri only allows select sections to be duplicated within one file.
+        When split across multiple files, we can use `include` to merge them safely.
+
+        Can be used impurely and with files commited to `/nix/store`.
```
llakala: We don't need a mention of this

modules/niri.nix
```diff
@@ -16,6 +16,23 @@
       '';
     };
 
+    sourceFiles = {
```
llakala: What if we just replaced `configFile` with this?

modules/niri.nix
```diff
@@ -16,6 +16,23 @@
...
+      type = types.listOf types.pathLike;
+      description = ''
+        A list of kdl files to import into the wrapped package's config.
+
+        Niri only allows select sections to be duplicated within one file.
```
llakala: This also doesn't need to be explained.

modules/niri.nix
```diff
@@ -16,6 +16,23 @@
       '';
     };
 
+    sourceFiles = {
+      type = types.listOf types.pathLike;
+      description = ''
+        A list of kdl files to import into the wrapped package's config.
```
llakala: How about "a list of `.kdl` files that are included in the wrapped package's config".

modules/niri.nix
```diff
@@ -25,7 +42,22 @@
...
-    if options ? configFile then
+    let
+      inherit (builtins) concatStringsSep;
+      inherit (inputs.nixpkgs.pkgs) writeText;
+      inherit (inputs.nixpkgs.lib) optionals;
+
+      assembledConfig = concatStringsSep "\n" (
+        (optionals (options ? configFile) [ ''include "${options.configFile}"'' ])
+        ++ (optionals (options ? sourceFiles) (
+          map (el: ''include optional=true "${el}"'') options.sourceFiles
```
llakala: I don't get why this facilitates impurity. Even with an impure path, it's just a string, and the file should still expect the path to exist.

llakala (changes requested):
cc @Coca162 as the module maintainer as well.

modules/niri.nix
bivsk: Perhaps I worded it wrong. In my use, I am providing a path to a config file that noctalia can generate. This doesn't always exist, so I set optional=true so it doesn't warn/error.

modules/niri.nix
bivsk: I am very happy going this route if it doesnt cause too much breakage. Will see what coca thinks.

bivsk:
@Coca162 what are your thoughts on changing `configFile` to optionally allow for a list of files? 

Logic should be similar to what I'm doing now, so happy to make the change if you agree it makes sense.

Coca162:
I feel ambivalent about this because I don't know if it makes sense to be a abstraction adios-wrappers provides to user ootb as it may not suit the needs of users who'd want it (you can see this issue with how `optional=true` while needed in your case might be a problem for others who would want to know if a file doesn't exist). However its not too complicated for me to be worried about maintaining even if I'm not using it.

If you wanted to do this without changing `impl` you could add a `defaultFunc` on `configFile` that just sets it based on a injected `sourceFiles`, that way it wouldn't have to be upstreamed.

Trying to look at potentially some broader adios-wrapper philosophies does it make sense to have abstractions like this in upstream? We follow RFC 42 because it is a direct mapping of nix configuration to the program's configuration with no real abstraction between it, and the same generally follows with configuration files as well. I'm not sure if there is any guidelines there.

modules/niri.nix
llakala: If Noctalia doesn't generate it, it should avoid mutating in the first place. Couldn't it just send an empty list if the file isn't generated?

bivsk: Noctalia has its "template" generator thingy for changing color schemes on the fly at runtime. On first run, this file may not exist, however. By default it does not exist.

I.e. my noctalia module is not generating this file, but rather noctalia itself during runtime. 

If you think more users would expect an error than the flexibility of `optional=true` here, I'm open to removing it.

bivsk:
> I feel ambivalent about this because I don't know if it makes sense to be a abstraction adios-wrappers provides to user ootb as it may not suit the needs of users who'd want it (you can see this issue with how `optional=true` while needed in your case might be a problem for others who would want to know if a file doesn't exist). However its not too complicated for me to be worried about maintaining even if I'm not using it.

I tried doing this similarly to the nushell wrapper's `sourceFiles`. Regarding the `optional=true` bit, I was not 100% certain on this myself, but I can make either way work for me. No real preference.


> If you wanted to do this without changing `impl` you could add a `defaultFunc` on `configFile` that just sets it based on a injected `sourceFiles`, that way it wouldn't have to be upstreamed.

I suppose this is fine for my usecase as well. I did not know i could inject entire options.

> Trying to look at potentially some broader adios-wrapper philosophies does it make sense to have abstractions like this in upstream? We follow RFC 42 because it is a direct mapping of nix configuration to the program's configuration with no real abstraction between it, and the same generally follows with configuration files as well. I'm not sure if there is any guidelines there.

To my understanding, RFC42 is reserved for `options.settings`. Although, I completely agree that there is ambiguity with languages where RFC42-style mappings do not work.  It seems other modules here have provided some levels of abstraction, however where the line is drawn - I am not sure.

bivsk:
Rebased and fixed nitpicks

llakala:
My take - this is a good test of how we should handle generalizability vs. convenience.

You could absolutely just replace `configFile` with logic to include mutators via an injection. However, I think that would be more painful to put on mutator-inclined people than making non-mutator-inclined people use a singleton list.

If I wanted to transform the current module into one that works with mutations, I would have to do:
```nix
# wrappers/niri/default.nix
{
  options.configFile = {
    mutators = [ "/foo" "/bar" ];
    mergeFunc =
      { mutators, inputs }:
      inputs.nixpkgs.pkgs.writeText "config" (
        builtins.concatStringsSep "/n" (
          map (file: '''include "${file}"'') mutators
        )
      );
  };
}
```
This is relatively hellish, compared to what people who _don't_ use mutators would have to do:
```nix
# wrappers/niri/default.nix
{
  options.configFiles.default = [
    ./config.kdl
  ];
}
```
I agree with making users handle custom abstractions when the abstraction is relatively painless. But I think this is a case where changing configFile -> configFiles  (or `sourcedFiles`, take your pick) is more worth it.

modules/niri.nix
llakala: I think we should remove it. PI'd like mutators to more easily support the ability to _stop_ mutating if they change their mind. But that requires more design work.

bivsk:
> I agree with making users handle custom abstractions when the abstraction is relatively painless. But I think this is a case where changing configFile -> configFiles (or `sourcedFiles`, take your pick) is more worth it.

Okay, made some changes. Let me know if I understood correctly.

@Coca162 let me know if this looks alright to you.

modules/niri.nix
```diff
@@ -1,19 +1,22 @@
...
   options = {
-    configFile = {
-      type = types.pathLike;
+    configFiles = {
+      type = types.listOf types.pathLike;
       description = ''
-        `config.kdl` file to be injected into the wrapped package.
+        A list of `.kdl` files that are included in the wrapped package's config.
+
+        If used along with the `configFile` option, values here take precedence.
```
Coca162: Line is now outdated

Coca162:
Seems all good code wise, would be nice if there was some way to indicate to users updating that the option no longer exists but idk if there is idiomatic way of doing that in adios (deprecation warning maybe?)

modules/niri.nix
bivsk: Whoops, thanks. Fixed.

llakala:
> Seems all good code wise, would be nice if there was some way to indicate to users updating that the option no longer exists but idk if there is idiomatic way of doing that in adios (deprecation warning maybe?)

Not yet.

Coca162 (approved):

PR #96, opened 2026-07-29
CONTRIBUTING.md: explain impurity in detail

### LINKED ISSUE #95: Document difference in `toString` between flakes and non-flakes
I believe it would be useful to add a note to the documentation explaining how `toString` works differently in flakes and non-flakes.
 
When using flakes, the entire flake's source tree is copied to the store. Any paths within our flake's directory referenced via `toString` are resolved using this store path. Thus, using `toString ./myConfig.toml` forces users to rebuild to actually apply any updates. 

Since much of the README discusses the benefits of using "impure" paths, I think this can confuse many flake users (it did with me).

I'm just now beginning to learn about these small differences, so I'm not sure on the correct verbiage to use, or I'd submit a PR myself. Also, please lmk if I'm misunderstanding this lol

llakala:
Fixes #95. @bivsk, let me know if this clarifies it as you hoped it would!
## Checklist

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -4,15 +4,41 @@ More modules and options are always welcome. However, they're expected to follow
 
 ## RFC42 and impure options
 
-Each option should aim to have an RFC42 variant and an impure path variant.
+Each config file that's used in a wrapper should have two options provided for handling it.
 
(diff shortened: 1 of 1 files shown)


bivsk:
Yes, this is perfect. Thanks!

PR #97, opened 2026-07-30
modules/fuzzel: remove `preWrap`

Coca162:
This check is both broken and also does not support impure paths.

It could be fixed and also check if the config is in the store but generally adios-wrapper's do not do any config validation anyways.

## Checklist

### DIFF
--- a/modules/fuzzel.nix
+++ b/modules/fuzzel.nix
@@ -79,7 +79,6 @@
   impl =
     { options, inputs }:
     let
-      inherit (builtins) head;
       inherit (inputs.nixpkgs.lib) optionals;
       inherit (inputs.nixpkgs.pkgs) formats;
       generator = formats.ini {};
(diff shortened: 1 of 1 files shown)


llakala:
cc @aabush64 

Definitely looks broken. But why wouldn't it work with impure paths? I mean, it wouldn't _check_ the impure path, but I don't think it would break impurity, which I consider more valuable.

Coca162:
> but I don't think it would break impurity

I haven't actually checked it but fuzzel does error if the config path does not exist, which presumably in the nix sandbox it wouldn't. I will check it now tho

Coca162:
yeah (the path exists and works if its in the nix store):
```
error: builder for '/nix/store/52f8l271wg551cbybh667wr8zf4gms9s-fuzzel-wrapped.drv' failed with exit code 1;
       last 1 log lines:
       >  err: config.c:1747: /home/coca/Desktop/nix-config/wrappers/fuzzel.ini: failed to open: No such file or directory (2)
       For full logs, run:
```

llakala:
Gotcha, sandboxing moment. Makes sense, I'll wait on Mango to approve.

llakala:
Mango's been inactive. LGTM.

PR #98, opened 2026-08-01
wrappers: use XOR for assertions where it's trivial

llakala:
cc: @aabush64 @Coca162 

By asserting that the values aren't equal, we can guarantee that only one of them is set, and ignore the possiblity that none are set. I only applied this to modules that only define two disjoint options (think `settings` and `configFile`. Any modules that had more options (like `theme`) could hypothetically be called with that other option, and have the primary option ignored. We can think about those in the future.

## Checklist

- [x] I tested my wrapper to make sure it functioned (I got the same hash for my devshell)

### DIFF
--- a/modules/btop.nix
+++ b/modules/btop.nix
@@ -62,19 +62,17 @@
         } " = ";
       };
     in
-    assert !(options ? settings && options ? configFile);
+    assert options ? settings != options ? configFile;
     inputs.mkWrapper {
       inherit (options) package;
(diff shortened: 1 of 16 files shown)


modules/btop.nix
Coca162, line 47: `optionals` should not be needed anymore

Coca162 (commented):
Looks good to me otherwise

PR #100, opened 2026-08-02
modules/lazygit: init

Coca162:
## Checklist

### DIFF
--- a/modules/lazygit.nix
+++ b/modules/lazygit.nix
@@ -1,0 +1,63 @@
+{ types, ... }:
+{
+  inputs = {
+    mkWrapper.from = { parent }: parent.mkWrapper;
+    nixpkgs.from = { parent }: parent.nixpkgs;
+  };
+
+  options = {
+    settings = {
+      type = types.attrs;
+      description = ''
+        Settings to be injected into the wrapped package's `config.yml`.
+
+        See the documentation for valid options:
+        https://github.com/jesseduffield/lazygit/blob/master/docs/Config.md
+
+        Disjoint with the `configFile` option.
+      '';
+    };
+    configFile = {
+      type = types.pathLike;
+      description = ''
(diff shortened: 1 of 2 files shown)


modules/lazygit.nix
```diff
@@ -0,0 +1,63 @@
...
+  };
+
+  options = {
+    settings = {
+      type = types.attrs;
+      description = ''
+        Settings to be injected into the wrapped package's `config.yml`.
+
+        See the documentation for valid options:
+        https://github.com/jesseduffield/lazygit/blob/master/docs/Config.md
```
Squawkykaka, line 15: Documentation should be done as markdown links rather than just a link.

`[documentation](https://documentation.link)`

modules/lazygit.nix
```diff
@@ -0,0 +1,63 @@
...
+        Disjoint with the `configFile` option.
+      '';
+    };
+    configFile = {
+      type = types.pathLike;
+      description = ''
+        `config.yml` file to be injected into the wrapped package.
+
+        See the documentation for syntax and valid options:
+        https://github.com/jesseduffield/lazygit/blob/master/docs/Config.md
```
Squawkykaka, line 26: see above

Squawkykaka (commented):
Seems good apart from these minor nits

modules/lazygit.nix
Coca162: hmmm

what about `<https://documentation.link>`

Coca162: actually links are already autodetected for the site so that wouldn't be better, bagh

llakala: Yes but not yet - I'm gonna apply that in a treewide, so this can stay as is for now.

llakala:
Forgot about this - LGTM.

PR #104, opened 2026-08-19
neovim: init wrapper

Squawkykaka:
## Checklist
A neovim wrapper yay, i have copied a lot of the code from mnw, so i gave some attribution at the top as a comment

This is slightly different than mnw, as in it does not have the `providers` or `devPlugins`. Dev plugins are instead detected automatically if a startPlugin is a string path.

I also chose not to add some attributes, but if they are wanted i will add them in.


Squawkykaka:
Implemented @llakala 's idea to have `luaContents` and `initLua` disjoint
also removed the VimL file handling

modules/neovim.nix
```diff
@@ -0,0 +1,299 @@
...
+          inherit (pkgs.vimPlugins) telescope;
+          # dev mode
+          myconfig = toString ./config
+        }
+      '';
+    };
+
+    optPlugins = {
+      type = types.attrsOf (
+        types.union [
```
llakala: pathLike is defined as
```nix
  pathLike = types.new {
    name = "pathLike";
    verify = v: isPath v || isDerivation v || isString v;
  };
```
So not necessary.

modules/neovim.nix
```diff
@@ -0,0 +1,299 @@
...
+
+      isTreesitter = p: p.isTreesitterGrammar or false || p.isTreesitterQuery or false;
+      optPlugins =
+        if options ? optPlugins then
+          filterAttrs (_: v: v != null && !isTreesitter v) options.optPlugins
+        else
+          { };
+      startPlugins =
+        let
+          allTreesitter = builtins.filter isTreesitter (
```
llakala: Inherit `filter`.

modules/neovim.nix
```diff
@@ -0,0 +1,299 @@
...
+      example = ''
+        with inputs.nixpkgs; [
+          pkgs.rg
+          pkgs.fzf
+        ]
+      '';
+    };
+
+    extraLuaPackages = {
+      type = types.any;
```
llakala: Still `types.any`.

modules/neovim.nix
```diff
@@ -0,0 +1,299 @@
...
+      default = _: [ ];
+      description = ''
+        A list of extra needed lua packages.
```
llakala: Not a list, is it?

modules/neovim.nix
```diff
@@ -0,0 +1,299 @@
...
+              nvim-treesitter-grammars = symlinkJoin {
+                name = "nvim-treesitter-grammars";
+                paths = allTreesitter;
+              };
+            }
+          else
+            { }
+        );
+      devPlugins =
+        if options ? startPlugins then filterAttrs (_: v: isString v) options.startPlugins else { };
```
llakala: Nit: can be `_: isString`

modules/neovim/default.nix
```diff
@@ -0,0 +1,299 @@
...
+
+      generatedInitLua =
+        let
+          luaEnv = options.package.lua.withPackages options.extraLuaPackages;
+          inherit (options.package.lua.pkgs) luaLib;
+
+          sourceLua = if options ? initLua then "dofile('${options.initLua}')" else options.luaContents;
+        in
+        writeText "init.lua" /* lua */ ''
+          adioswrappers = { configDir = "$out" }
```
llakala: Nit - we use `adios-wrappers` consistently.

modules/neovim.nix
```diff
@@ -0,0 +1,299 @@
...
+          package.cpath = "${luaLib.genLuaCPathAbsStr luaEnv};$LUA_CPATH" .. package.cpath
+
+          ${sourceLua}
+        '';
+
+      configDir = stdenvNoCC.mkDerivation {
+        name = "neovim-configDir";
+        nativeBuildInputs = [ envsubst ];
+        __structuredAttrs = true;
+        preferLocalBuilds = true;
```
llakala: Should be `preferLocalBuild`.

modules/neovim.nix
```diff
@@ -0,0 +1,299 @@
...
+      '';
+      example = ''
+        ps: [ ps.jsregexp ]
+      '';
+    };
+
+    startPlugins = {
+      type = types.attrsOf types.pathLike;
+      description = ''
+        The list of plugins to add to the nvim config, and run at startup.
```
llakala: This is an attrset, not a list.

modules/neovim.nix
```diff
@@ -0,0 +1,299 @@
...
+        getVersion
+        makeBinPath
+        mapAttrsToList
+        getExe
+        filterAttrs
+        escapeShellArgs
+        ;
+      inherit (builtins) concatStringsSep isString attrValues;
+
+      isTreesitter = p: p.isTreesitterGrammar or false || p.isTreesitterQuery or false;
```
llakala: I think we have to do a more careful dance to get treesitter to work. For `my nvim-treesitter.withPlugins` to work, it needs to be in `start`, so the dependencies get auto-checked.

I think before we work with `startPlugins`, we need to recurse into them and find any dependencies. This is done [here](https://github.com/Gerg-L/mnw/blob/0151d9fa87992cbeb67c86616fb9170dcec830b6/modules/options.nix#L363) in mnw.

modules/neovim.nix
```diff
@@ -0,0 +1,299 @@
...
+      type = types.string;
+      description = "lua config text to load at startup. Disjoint with `initLua`";
+      example = ''
+        require("myConfig")
+      '';
+    };
+
+    aliases = {
+      type = types.listOf types.string;
+      default = [
```
llakala: I'd make this empty by default, as it is in mnw.

modules/neovim.nix
```diff
@@ -0,0 +1,299 @@
...
+          { };
+      startPlugins =
+        let
+          allTreesitter = builtins.filter isTreesitter (
+            attrValues (options.startPlugins or { }) ++ attrValues (options.optPlugins or { })
+          );
+        in
+        (filterAttrs (_: v: v != null && !isTreesitter v && !isString v) (options.startPlugins or { }))
+        // (
+          if allTreesitter != [ ] then
```
llakala: Seeing this makes me think we should move to an option for treesitter specifically, along the lines of
```nix
treesitterPackage = {
  type = types.package;
  description = ''
    The nvim-treesitter package to be used.

    This should also include the grammars as dependencies, which can be done via either 
    `nvim-treesitter.withAllGrammars` or `nvim-treesitter.withPlugins (p: [ p.foo p.bar ])`.
  '';
  example = ''pkgs.vimPlugins.nvim-treesitter.withAllGrammars';
}

File: modules/neovim.nix [hunk 12]
```diff
@@ -0,0 +1,299 @@
...
+              rmdir $path
+            fi
+          done
+        '';
+      };
+
+      wrapperArgsStr = escapeShellArgs [
+        "--add-flags"
+        "--cmd \"lua vim.opt.packpath:prepend('${configDir}'); vim.opt.runtimepath:prepend('${configDir}'); ${
+          if devPlugins != { } then
```
llakala: Maybe we need more intelligent logic here, to allow easily enabling/disabling a dev mode. Otherwise the impure mode will always be enabled. Granted, this is an issue for all our wrappers, and I haven't thought of a good way to clean this up. An input to `mkWrapper` that toggles global purity on/off, maybe.

For now, how about:
1. devPlugins option just like mnw, with this type:
```nix
attrsOf (struct "devPlugin" {
  impure = types.string;
  pure = types.union [ types.package types.derivation ];
})
```
2. An `impure` option (defaulting to false) that determines whether to read the `.pure` or `.impure` attribute of  each dev plugin. When `impure == false`, dev plugins are included in startPlugins. When `impure == true`, dev plugins get loaded with this logic.

3. With this, it's easy to enable the dev mode - just `tree.modules.neovim { impure = true; }`. We can mention this in the description of the `impure` attribute.

modules/neovim.nix
llakala: Not a list.

modules/neovim.nix
```diff
@@ -0,0 +1,299 @@
...
+
+    optPlugins = {
+      type = types.attrsOf (
+        types.union [
+          types.path
+          types.derivation
+        ]
+      );
+      description = ''
+        A list of nvim plugins to load when needed.
```
llakala: Not a list.

llakala (changes requested):

modules/neovim/default.nix
Squawkykaka: Lua does not support `-` inside variable names

llakala: Gross. Add a comment?

modules/neovim.nix
Squawkykaka: The bash code used for config file generation doesnt support copying from string paths, which is why i made the custom type

llakala:
Committed .direnv.

Squawkykaka:
Apologies, fixed

modules/neovim.nix
```diff
@@ -0,0 +1,366 @@
...
+        A attrset of plugins to add to the nvim config, and run at startup.
+
+        If a path is a string, it will be loaded at runtime instead of at build
+        time allowing for hot-reloading.
+      '';
+      example = ''
+        {
+          inherit (pkgs.vimPlugins) telescope;
+          # dev mode
+          myconfig = toString ./config
```
Ladas552: shouldn't it have `;` at the end
`myconfig = toString ./config;`

Squawkykaka: Fixed

modules/neovim.nix
```diff
@@ -0,0 +1,375 @@
...
+  };
+
+  options = {
+    package = {
+      type = types.derivation;
+      defaultFunc = { inputs }: inputs.nixpkgs.pkgs.neovim-unwrapped;
+      description = "The neovim package to be wrapped.";
+    };
+
+    initLua = {
```
llakala: I assume this probably needs to be in the store? Ideally, the type would reflect, or it would at least be documented.

modules/mkWrapper/default.nix
```diff
@@ -134,12 +139,12 @@ in {
...
       buildInputs = [ makeBinaryWrapper ];
       paths =
-        if options.extraPaths == [] then
+        if options.extraPaths == [ ] then
           [ "${options.package}" ]
         else
           map (path: "${path}") ([ options.package ] ++ options.extraPaths);
       meta.mainProgram = options.binaryName;
-      passthru = options.package.passthru or {};
+      passthru = if options ? passthru then options.passthru else options.package.passthru or { };
```
llakala: We should merge together `options.package.passthru // options.passthru` if we do this. 

Technically, this doesn't need to be added as an option - we can just `//` the package, like this:
```nix
inputs.mkWrapper {
  package = options.package // {
    passthru = (options.package.passthru or {}) // {
    a = b;
  };
}
```
Just up to you whether thats too annoying. Providing an option might be nice.

Squawkykaka: Changed to use that method

modules/neovim/default.nix
```diff
@@ -0,0 +1,245 @@
...
+          inherit pluginDir;
+          config = options // {
+            startPlugins = transformedStartPlugins;
+          };
+        };
+      };
+      environment.VIMINIT = "source $out/share/nvim/init.lua";
+      symlinks = {
+        "$out/share/nvim/init.lua" = generatedInitLua;
+        "$out/share/nvim/plugins" = "${pluginDir}/pack/plugins";
```
Squawkykaka, line 232: Is this really necessary? i just build `wrappers.neovim.drv.configDir` to see the generated config, i dont think it needs to be symlinked in.

PR #105, opened 2026-08-23
site: homepage face lift

yaaaarn:
## Checklist


llakala (changes requested):
Thanks for this!

Looks good to me. Could we change the ordering, so, so it's sorted in column-major order rather than row-major? I'd find that easier to read.

PR #106, opened 2026-08-23
site: module page improvements

yaaaarn:
## Checklist

### DIFF
--- a/site/src/pages/modules/[id].astro
+++ b/site/src/pages/modules/[id].astro
@@ -6,6 +6,9 @@ import { jsonValueToNix, type JSONType } from "../../lib/nix";
 import { marked } from "../../lib/highlight";
 import Definition from "../../components/Definition.astro";
 
+import fs from 'fs';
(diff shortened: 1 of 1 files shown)


site/src/pages/modules/[id].astro
```diff
@@ -32,8 +32,14 @@ function stringifyExample(example: JSONType): string {
     <title>{id} - adios-wrappers</title>
   </Fragment>
 
-  <h1>{id}</h1>
-  <h2>Options</h2>
+  <h1 class="id">
+    {id}
+
+    <a href={`https://github.com/llakala/adios-wrappers/tree/main/modules/${id}.nix`}>Source file</a>
```
Comment (llakala) [inline site/src/pages/modules/[id].astro]: Nit - this isn't _always_ true. mkWrapper is actually in modules/mkWrapper/default.nix. Maybe we could check if the URL exists at site build time, and if not, use the default.nix one?

llakala (changes requested):
Tiny thing.

PR #108, opened 2026-08-29
modules/rmpc: init

itsyunaya:
## Checklist

### DIFF
--- a/modules/rmpc.nix
+++ b/modules/rmpc.nix
@@ -1,0 +1,81 @@
+{ types, ... }: {
+  inputs = {
+    nixpkgs.from = { parent }: parent.nixpkgs;
+    mkWrapper.from = { parent }: parent.mkWrapper;
+  };
+
+  options = {
+    configContents = {
+      type = types.string;
+      description = ''
+        Configuration to be injected into the wrapped package's `config.ron`.
+
+        See the [documentation](https://rmpc.mierak.dev/configuration/) on valid options.
+
+        Disjoint with the `configFile` option.
+      '';
+    };
+    configFile = {
+      type = types.pathLike;
+      description = ''
+        `config.ron` file to be injected into the wrapped package.
+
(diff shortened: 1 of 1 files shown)


modules/rmpc.nix
```diff
@@ -0,0 +1,92 @@
...
+      description = ''
+        `config.ron` file to be injected into the wrapped package.
+
+        See the [documentation](https://rmpc.mierak.dev/configuration/) on valid options.
+
+        Disjoint with the `configContents` option.
+      '';
+    };
+
+    theme = {
```
llakala: I wonder if it would make more sense to have a  `themes` attrset, of type `atrrsOf<union<attrs,pathLike>>`. This would generate the themes directort, and the user could then select any theme they like in their config. If you want an example of how to generate a directory like this, check out the Yazi wrapper.

llakala (changes requested):

modules/rmpc.nix
itsyunaya: Having an attrset of themes does sound more user-friendly, but it would, at least in my opinion, still make sense to separate ones imported from a path from inline ones. Mainly so the structure of the other modules is kept up, but also because it makes it less messy to define. There's also unfortunately no `.ron` generator in Nixpkgs, so inline themes will have to be defined as a string.

llakala: > There's also unfortunately no .ron generator in Nixpkgs, so inline themes will have to be defined as a string.

Yeah, that's fine. Just wrote attrs on reflex.

> Having an attrset of themes does sound more user-friendly, but it would, at least in my opinion, still make sense to separate ones imported from a path from inline ones. Mainly so the structure of the other modules is kept up, but also because it makes it less messy to define.

Yeah, it would be an issue with impure paths, since you can't tell the string apart. How about we just only allow the path variant for themes for simplicity? Then it's:
```nix
{
  themes = {
    light = ./light.ron;
    dark = ./dark.ron;
  };
}
```

llakala (approved):

PR #109, opened 2026-08-30
modules/anyrun: init

itsyunaya:
I'm a bit unsure about the `pluginPaths` and `pluginConfigFiles` implementation on this one, any feedback would be greatly appreciated.

## Checklist

### DIFF
--- a/modules/anyrun.nix
+++ b/modules/anyrun.nix
@@ -1,0 +1,110 @@
+{ types, ... }: {
+  inputs = {
+    mkWrapper.from = { parent }: parent.mkWrapper;
+    nixpkgs.from = { parent }: parent.nixpkgs;
+  };
+
+  options = {
+    configFiles = {
+      type = types.attrsOf types.pathLike;
+      description = ''
+        Anyrun and plugin config files to be injected into the wrapped package.
+
+        See the [config example](https://github.com/anyrun-org/anyrun/blob/master/examples/config.ron) for valid anyrun options and the [plugins directory](https://github.com/anyrun-org/anyrun/tree/master/plugins) for official plugins.
+      '';
+      example = {
+        "config.ron" = ./config.ron;
+        "shell.ron" = "/home/user/shell.ron";
+      };
+    };
+
+    cssContents = {
+      type = types.string;
+      description = ''
+        CSS to be injected into the wrapped package.
+
+        Refer to the [example stylesheet](https://github.com/anyrun-org/anyrun/blob/master/anyrun/res/style.css) for possible configuration.
+
+        Disjoint with the `cssFile` option.
+      '';
+    };
+    cssFile = {
+      type = types.pathLike;
+      description = ''
+        `style.css` file to be injected into the wrapped package.
+
+        Refer to the [example stylesheet](https://github.com/anyrun-org/anyrun/blob/master/anyrun/res/style.css) for possible configuration.
+
+        Disjoint with the `cssContents` option.
+      '';
+    };
+
+    pluginPaths = {
+      type = types.listOf types.pathLike;
+      description = ''
+        Paths of anyrun plugins to be installed into the plugins directory.
+
+        Must point to a plugin's shared object file (`.so`).
+      '';
+      example = [
+        "\${inputs.nixpkgs.pkgs.anyrun}/lib/libapplications.so"
+        "/home/user/anyrun-plugins/libshell.so"
+      ];
+    };
+
+    package = {
+      type = types.derivation;
+      defaultFunc = { inputs }: inputs.nixpkgs.pkgs.anyrun;
+      description = "The anyrun package to be wrapped.";
+    };
+  };
+
+  impl =
+    { options, inputs }:
+    let
+      inherit (inputs.nixpkgs.pkgs) writeText;
+      inherit (builtins) attrNames baseNameOf listToAttrs unsafeDiscardStringContext;
+      optionalAttrs = cond: attrs: if cond then attrs else {};
+    in
+    assert !(options ? cssContents && options ? cssFile);
+    inputs.mkWrapper {
+      inherit (options) package;
+      symlinks = {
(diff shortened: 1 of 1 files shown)


modules/anyrun.nix
```diff
@@ -0,0 +1,131 @@
...
+      type = types.pathLike;
+      description = ''
+        `config.ron` file to be injected into the wrapped package.
+
+        See the [config example](https://github.com/anyrun-org/anyrun/blob/master/examples/config.ron) for valid options.
+
+        Disjoint with the `configContents` option.
+      '';
+    };
+
```
Squawkykaka, line 20: There shouldnt be any spaces between option definitioons

modules/anyrun.nix
```diff
@@ -0,0 +1,131 @@
...
+  impl =
+    { options, inputs }:
+    let
+      inherit (inputs.nixpkgs.pkgs) writeText;
+      inherit (builtins) baseNameOf listToAttrs unsafeDiscardStringContext;
+      optionalAttrs = cond: attrs: if cond then attrs else {};
+    in
+    assert !(options ? configContents && options ? configFile);
+    assert !(options ? cssContents && options ? cssFile);
```
Squawkykaka: Options such as configFile make sense to always be set usually, so XOR is used instead which makes handing nicer in the code

instead of
```nix
assert !(options ? configContents && options ? configFile);
```

youd have
```nix
assert options ? configContents != options ? configFile;
```

modules/anyrun.nix
```diff
@@ -0,0 +1,131 @@
...
+    pluginPaths = {
+      type = types.listOf types.pathLike;
+      description = ''
+        Paths of anyrun plugins to be installed into the plugins directory.
+
+        Must point to a plugin's shared object file (`.so`).
+      '';
+    };
+    pluginConfigFiles = {
+      type = types.listOf types.pathLike;
```
Squawkykaka: I would probably use `types.attrsOf` instead, so that the name of the config could be handled a bit more ergonomically, this would be the same for `pluginPaths`

```suggestion
      type = types.attrsOf types.pathLike;
```

modules/anyrun.nix
```diff
@@ -0,0 +1,131 @@
...
+        )
+      )
+      // optionalAttrs (options ? pluginPaths) (
+        listToAttrs (
+          map (plugin: {
+            name = "$out/anyrun/plugins/${unsafeDiscardStringContext (baseNameOf plugin)}";
+            value = plugin;
+          }) options.pluginPaths
+        )
+      );
```
Squawkykaka: This could then be simplified with the change of the option type to

```suggestion
      // (if options ? pluginConfigFiles then options.pluginConfigFiles else {})
      // (if options ? pluginPaths then options.pluginPaths else {});
```

modules/anyrun.nix
```diff
@@ -0,0 +1,131 @@
...
+
+      environment = {
+        "XDG_CONFIG_HOME" = "$out";
```
Squawkykaka: Is there a application-specific environment variable for the config? we perfer using those over `XDG_CONFIG_HOME` whenever possible

Also the quotes around XDG_CONFIG_HOME are not necicary
```suggestion
        XDG_CONFIG_HOME = "$out";
```

Squawkykaka (changes requested):

Squawkykaka:
Awesome job for a first module anyways

modules/anyrun.nix
llakala: You still need to set the name to $out/anyrun.

modules/anyrun.nix
llakala: Also, don't these two options end up in the same place? Not sure if there's any reason we need both of them.

itsyunaya: Anyrun is a bit weird, in the sense that plugins go into `$XDG_CONFIG_HOME/anyrun/plugins/`, while the config files for them stay in `$XDG_CONFIG_HOME/anyrun/`, so these do have to stay separated. Although this means that technically all config files (the main one and any for plugins) could be set via one option of type `attrsOf<pathLike>` to simplify things.

I also did think of making the plugin paths `attsOf<pathLike>` too, but then we'd end up with something like this, and I'm not sure if that isn't a bit redundant, especially since I don't think you're allowed to change the name of the plugin files:
```nix
pluginPaths = {
  "libshell.so" = "${pkgs.anyrun}/lib/libshell.so";
  "librink.so" = "${pkgs.anyrun}/lib/librink.so";
  # and so on
};
```

modules/anyrun.nix
itsyunaya: Unfortunately it only reliably checks for `~/.config/anyrun` and `$XDG_CONFIG_HOME/anyrun`. It also searches other places, but that's apparently legacy behaviour and should therefore not be relied on.

modules/anyrun.nix
```diff
@@ -0,0 +1,102 @@
+{ types, ... }: {
+  inputs = {
+    mkWrapper.from = { parent }: parent.mkWrapper;
+    nixpkgs.from = { parent }: parent.nixpkgs;
+  };
+
+  options = {
+    configFiles = {
```
llakala, line 8: Could we get an example of this option and the pluginPaths option?

llakala: Ideally demoing both local and derivation paths.

llakala (changes requested):
Only one thing.

itsyunaya:
Examples are done now. The string interpolation in the `pluginPaths` one is escaped because without it, LSPs would mark it as an error (`inputs` isn't defined). I wasn't sure how to do it otherwise without writing out the entire function. I hope that's not a problem for docs generation.

llakala:
Sometimes we just use a string and _pretend_ it's an attrset example. You lose syntax highlighting that way, but we lose it here too.

I see a few small cleanups, but I'll just add them after merging.

PR #114, opened 2026-09-03
modules/ghostty: init

bivsk:
## Checklist

### DIFF
--- a/modules/ghostty.nix
+++ b/modules/ghostty.nix
@@ -1,0 +1,81 @@
+{ types, ... }:
+{
+  inputs = {
+    mkWrapper.from = { parent }: parent.mkWrapper;
+    nixpkgs.from = { parent }: parent.nixpkgs;
+  };
+
+  options = {
+    settings = {
+      type = types.attrs;
+      description = ''
+        Options to be injected into the wrapped package's `config.ghostty`.
+
+        See the ghostty [documentation](https://ghostty.org/docs/config).
+
+        Disjoint with the `configFile` option.
+      '';
+      example = {
+        cursor-color = "ffffff";
+        cursor-text = "000000";
+        background = "272822";
+        foreground = "ffffff";
+      };
+    };
+    configFile = {
+      type = types.pathLike;
+      description = ''
+        A list of `config.ghostty` files to be injected into the wrapped package.
+
+        See the ghostty [documentation](https://ghostty.org/docs/config).
+
+        Disjoint with the `settings` option.
+      '';
+    };
+
+    package = {
+      type = types.derivation;
+      defaultFunc = { inputs }: inputs.nixpkgs.pkgs.ghostty;
+      description = "The ghostty package to be wrapped.";
+    };
+  };
+
+  impl =
+    { options, inputs }:
+    let
+      inherit (inputs.nixpkgs.pkgs) formats;
+      generator = formats.keyValue {
+        listsAsDuplicateKeys = true;
+      };
+    in
+    assert (options ? settings != options ? configFile);
+    inputs.mkWrapper {
+      inherit (options) package;
+      # Ensure ghostty.service uses the wrapped package.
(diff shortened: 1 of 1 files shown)


bivsk:
some kind of helper to edit store path references in systemd service files would be quite useful

modules/ghostty.nix
```diff
@@ -0,0 +1,92 @@
...
+          head options.configFiles
+        else
+          writeText "config.ghostty" (
+            concatStringSep "\n" (map (file: ''config-file "${file}"'') options.configFiles)
+          );
+    in
+    assert !(options ? settings && options ? configFiles);
+    inputs.mkWrapper {
+      inherit (options) package;
+      postWrap = ''
```
llakala, line 56: I get what this does, but could we have a comment explaining it anyways?

modules/ghostty.nix
```diff
@@ -0,0 +1,92 @@
...
+        cp $out/share/systemd/user/app-com.mitchellh.ghostty.service app-com.mitchellh.ghostty.service
```
llakala, line 57: Do we need to cp from the original location, just to make a change and put it back? Why not just substitute in the original location?

modules/ghostty.nix
```diff
@@ -0,0 +1,92 @@
...
+    };
+  };
+
+  impl =
+    { options, inputs }:
+    let
+      inherit (builtins) concatStringSep head length;
+      inherit (inputs.nixpkgs.pkgs) formats writeText;
+      inherit (inputs.nixpkgs.lib.lists) optionals;
+
```
llakala: Nit - no newlines here.

modules/ghostty.nix
```diff
@@ -0,0 +1,92 @@
...
+        Disjoint with the `configFiles` option.
+      '';
+      example = {
+        cursor-color = "ffffff";
+        cursor-text = "000000";
+        background = "272822";
+        foreground = "ffffff";
+      };
+      mergeFunc = adios.lib.merge.attrs.recursively;
```
llakala: Nit - are you planning on mutating these? I like to only set `mergeFunc` when there's someone who's found it useful to mutate a module. This keeps it as a "second-class citizen".

modules/ghostty.nix
```diff
@@ -0,0 +1,92 @@
...
+
+      configPath =
+        if length options.configFiles == 1 then
+          head options.configFiles
+        else
+          writeText "config.ghostty" (
+            concatStringSep "\n" (map (file: ''config-file "${file}"'') options.configFiles)
+          );
+    in
+    assert !(options ? settings && options ? configFiles);
```
llakala: Nit - if you change this to `options ? settings != options ? configFiles`, you can make the `flags` later unconditional.

llakala (changes requested):

modules/ghostty.nix
bivsk: Fair point. After some thinking, I also think changing `configFiles` to `configFile` makes more sense for the vast majority of use cases here.

modules/ghostty.nix
bivsk: Done. However, without a default value for either, the assertion triggers. No point using the wrapper in that case, but may be worth noting.

modules/ghostty.nix
bivsk: When trying to modify in place, we get a permission error: 
`chmod: changing permissions of '/nix/store/589mawfc3vy3chlfip8h40m5qnynfx5i-ghostty-wrapped/share/systemd/user/app-com.mitchellh.ghostty.service': Operation not permitted`

modules/ghostty.nix
llakala: Yeah, I've decided with a wrapper that sets no options, it makes sense to error.

llakala (approved):

PR #115, opened 2026-09-03
modules/nh: init

itsyunaya:
## Checklist

### DIFF
--- a/modules/nh.nix
+++ b/modules/nh.nix
@@ -1,0 +1,36 @@
+{ types, ... }: {
+  inputs = {
+    mkWrapper.from = { parent }: parent.mkWrapper;
+    nixpkgs.from = { parent }: parent.nixpkgs;
+  };
+
+  options = {
+    environment = {
+      type = types.attrs;
+      description = ''
+        Environment variables to be passed to the wrapped package.
+
+        See the [documentation](https://github.com/nix-community/nh#environment-variables) for valid options.
+      '';
+    };
+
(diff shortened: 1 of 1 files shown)


llakala:
Putting on my pedantic hat - if this is just setting `environment`, I'm not sure if we should provide a wrapper for it. This would be really easy for someone to do themselves.

itsyunaya:
The reason I made this a wrapper were mainly cross-platform concerns, since the way environment variables are set can differ between Nix-based platforms (e.g. nix-darwin doesn't have `environment.sessionVariables`) and this would make it slightly nicer to manage/easier to have in one place. Although if you think that doesn't warrant an entire wrapper, I can close the PR.

llakala:
To be clear, I agree it's nice to wrap - I just think people should do
```nix
# wrappers/nh.nix
{
  inputs.mkWrapper.from = { parent }: parent.mkWrapper;
  inputs.nixpkgs.from = { parent }: parent.nixpkgs;

  impl = { inputs }: inputs.mkWrapper {
    package = inputs.nixpkgs.pkgs.nh;
    environment = {
      SOME_VAR = true;
    };
  };
}
```
themselves.

I know we're a wrapper set - which means providing preconfigured wrappers for the user - but this is a case where the wrapper is so simple that I'd like it to teach people how to do it themselves. We could even have a section in the docs called "How to do it yourself" in the future, maybe separate from the contribution docs.

PR #116, opened 2026-09-03
modules/keychain: init

itsyunaya:
## Checklist

### DIFF
--- a/modules/keychain.nix
+++ b/modules/keychain.nix
@@ -1,0 +1,57 @@
+{ types, ... }:
+let
+  inherit (builtins) concatStringsSep;
+  command =
+    opts:
+    "${opts.package}/bin/keychain --eval --absolute --dir $XDG_RUNTIME_DIR/keychain ${
+      concatStringsSep " " (opts.flags ++ opts.keys)
+    }";
+in {
+  inputs = {
+    mkWrapper.from = { parent }: parent.mkWrapper;
+    nixpkgs.from = { parent }: parent.nixpkgs;
+  };
+
+  options = {
+    flags = {
+      type = types.listOf types.string;
+      default = [];
+      description = ''
+        Flags to be injected into the wrapped package.
+
+        See keychain's [readme](https://github.com/danielrobbins/keychain#2-embedded-documentation) on GitHub for valid options.
+      '';
+    };
+
+    keys = {
+      # Intentionally not listOf<pathLike>, since we don't want keys to be copied to the store
+      type = types.listOf types.string;
+      default = [];
+      description = "Paths to keys to be added to the wrapped package.";
+      example = [ "~/.ssh/id_ed25519" ];
+    };
+
+    package = {
+      type = types.derivation;
+      defaultFunc = { inputs }: inputs.nixpkgs.pkgs.keychain;
+      description = "The keychain package to wrap.";
+    };
+  };
+
+  mutations = {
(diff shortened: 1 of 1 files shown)


modules/keychain.nix
```diff
@@ -0,0 +1,67 @@
...
+  inherit (builtins) concatStringsSep;
+  xdgFlags = [
+    "--absolute"
+    "--dir"
+    "$XDG_RUNTIME_DIR/keychain"
+  ];
+  command =
+    opts:
+    "${opts.package}/bin/keychain --eval ${
+      concatStringsSep " " (xdgFlags ++ opts.flags)
```
llakala: Nit - this runs `concatStringsSep` with constant data, and also runs it twice. We could pre-apply it:
```nix
"${opts.package}/bin/keychain --eval --absolute --dir $XDG_RUNTIME_DIR/keychain ${
  concatStringsSep " " (opts.flags ++ opts.keys)
}";

File: modules/keychain.nix [hunk 2]
```diff
@@ -0,0 +1,67 @@
...
+    flags = {
+      type = types.listOf types.string;
+      description = ''
+        Flags to be injected into the wrapped package.
+
+        See keychain's [readme](https://github.com/danielrobbins/keychain#2-embedded-documentation) on GitHub for valid options.
+      '';
+    };
+
+    keys = {
```
llakala, line 26: `flags` and `keys` need a default, to prevent an error if one isn't set.

modules/keychain.nix
```diff
@@ -0,0 +1,67 @@
...
+      type = types.derivation;
+      defaultFunc = { inputs }: inputs.nixpkgs.pkgs.keychain;
+      description = "The keychain package to wrap.";
+    };
+  };
+
+  mutations = {
+    "/fish".interactiveShellInit = { options }: "eval (${command options})";
+    "/zsh".zshrc = { options }: ''eval "$(${command options})"'';
+    # Adapted from hm
```
llakala, line 44: Could we link the hm file we referenced, ideally with a permalink?

modules/keychain.nix
```diff
@@ -0,0 +1,67 @@
...
+    "/nushell".shellInit = { options }: ''
+      let keychain_shell_command = (SHELL=bash ${command options}| parse -r '(\w+)="?(.*?)"?; export \1' | transpose -ird)
+      if not ($keychain_shell_command|is-empty) {
+        $keychain_shell_command | load-env
+      }
+    '';
+  };
+
+  # In case someone using Bash or another shell ever wants to use this wrapper, they can with this
+  impl =
```
llakala: I'd remove this for now, just so people don't expect they'll have to install it. People can write a mutation themselves for something like bash, that reads from the package.

modules/keychain.nix
```diff
@@ -0,0 +1,67 @@
+{ types, ... }:
+let
+  inherit (builtins) concatStringsSep;
+  xdgFlags = [
+    "--absolute"
+    "--dir"
+    "$XDG_RUNTIME_DIR/keychain"
```
llakala: Is the idea that you'll place the real files in $XDG_RUNTIME_DIR?

llakala (changes requested):

modules/keychain.nix
itsyunaya: `$XDG_RUNTIME_DIR/keychain` just stores some ssh-agent related temporary files the program needs.

llakala (approved):

PR #126, opened 2026-09-15
modules/regreet: init

EpicEric:
Tested with the following config:

```nix
_: {
  options = {
    extraPackages.defaultFunc = { inputs }: [
      inputs.nixpkgs.pkgs.adwaita-icon-theme
      inputs.nixpkgs.pkgs.xcursor-pro
    ];
    fontsPackages.defaultFunc = { inputs }: [
      inputs.nixpkgs.pkgs.dejavu_fonts
    ];
    cssFile.default = ./tokyo-night.css;
    settings.default.GTK = {
      application_prefer_dark_theme = true;
      cursor_theme_name = "XCursor-Pro-Dark";
      font_name = "DejaVu Sans 16";
      icon_theme_name = "Adwaita";
    };
  };
}
```

## Checklist


modules/regreet.nix
```diff
@@ -0,0 +1,123 @@
...
+  };
+
+  options = {
+    settings = {
+      type = types.attrs;
+      description = ''
+        Settings to be injected into the wrapped package's configuration.
+
+        See the ReGreet docs for valid options:
+        https://github.com/rharish101/ReGreet/tree/main#configuration
```
llakala: Nit - we use markdown-style links now, since they render nice on the site. Should be something like `See the [ReGreet docs]()`

modules/regreet.nix
```diff
@@ -0,0 +1,123 @@
...
+
+        Disjoint with the `configFile` option.
+      '';
+      mergeFunc = adios.lib.merge.attrs.recursively;
```
llakala: Make sure to only include this if you're actually mutating this personally.

modules/regreet.nix
```diff
@@ -0,0 +1,123 @@
...
+      description = ''
+        GTK4 CSS file to be injected into the wrapped package.
+
+        See the ReGreet docs for details:
+        https://github.com/rharish101/ReGreet/tree/main#custom-css
+
+        Disjoint with the `cssContents` option.
+      '';
+    };
+    package = {
```
llakala: Nit - package option should come last.

modules/regreet.nix
```diff
@@ -0,0 +1,123 @@
...
+            null;
+        "$out/style.css" =
+          if options ? cssFile then
+            options.cssFile
+          else if options ? cssContents then
+            writeText "style.css" options.cssContents
+          else
+            null;
+        "$out/fonts.conf" =
+          if options.fontsPackages != [] then
```
llakala: I think this will error if you don't set `fontsPackages`. You'll need to either give it `default = []`, or check if `options ? fontsPackages` first. Former seems preferable.

modules/regreet.nix
```diff
@@ -0,0 +1,123 @@
...
+      type = types.listOf types.derivation;
+      description = ''
+        Packages to be automatically added to ReGreet's PATH.
+        Typically used for GTK4, icon, and cursor themes.
+      '';
+    };
+    fontsPackages = {
+      type = types.listOf types.derivation;
+      description = ''
+        Font packages looked up by ReGreet.
```
llakala, line 59: Got a docs link for this? If not, no worries.

llakala (changes requested):
Small stuff. Including the config you tested with is nice - might add that to the template.

modules/regreet.nix
EpicEric: Nice catch, I had [] as the default but ended up removing it from the option.

modules/regreet.nix
EpicEric: Not really, this is just what FontConfig uses.

EpicEric:
Actually, the PATH part should be wrong. I don't think these packages even expose binaries, and I imagine that GTK has its own mechanisms to fetch those packages. The only reason it should've worked on my machine was because I have these packages installed globally.

EpicEric:
The correct envvar for `extraPackages` is `XDG_DATA_DIRS`. Tested in a VM without global packages for cursor and icons:

image

modules/regreet.nix
```diff
@@ -0,0 +1,120 @@
...
+  };
+
+  impl =
+    { options, inputs }:
+    let
+      inherit (inputs.nixpkgs.lib) makeSearchPath optionals;
+      inherit (inputs.nixpkgs.pkgs) formats makeFontsConf writeText;
+      generator = formats.toml {};
+    in
+    assert !(options ? settings && options ? configFile);
```
llakala, line 76: Oh, one last thing. Assert cssContents and cssFile.

llakala (approved):
Thanks for the high-quality testing!

PR #127, opened 2026-09-15
modules/btop: add themes option

itsyunaya:
Allows adding custom theme files to the wrapper. Both custom and internal themes can be referenced by name when setting `color_theme` in the settings.

cc: @Coca162

## Checklist


modules/btop.nix
```diff
@@ -27,9 +27,10 @@
       '';
     };
 
-    # TODO: Themes are located in $out/share/btop/themes
-    # should provide a option here to be able to set them
-    # though I don't know what route would be best for that
+    themes = {
```
Coca162, line 30: This should be named `themesFiles` or something like it to indicate in the name the type more obviously (and for future proofing if other ways to specify are made)

Coca162 (commented):
I'm not a custom theme haver but this looks fine, my initial hesistation was because I was thinking about things like `themesDir` or also providing strings for the themes instead of files like in home-manager.

modules/btop.nix
llakala: themesFiles would strike me as a name of a list. Not sure what I'd name it.

itsyunaya: Technically this *would* also work as a list instead of an attrset:
```nix
themeFiles.default = [
  ./my-theme.theme
  ./second-theme.theme
]
```
The only thing we'd lose is the ability to give the themes names differing from their file names.

llakala: I mean, I like the aesthetics of that for your own files, but if you're vendoring a theme from someone else and it has a weird filename, that feels gross.

itsyunaya: Fair, in that case I think leaving the name as `themes` might be best, since that's what we're doing for this kind of option in other wrappers too.

Squawkykaka: Also having it as a list limits generation via stuff like writeFile, as the name will always contain the hash of the file, which is undesirable.

PR #128, opened 2026-09-18
modules/nnn: init

yaaaarn:
## Checklist


modules/nnn.nix
```diff
@@ -0,0 +1,77 @@
...
+        Disjoint with the `configContents` option.
+      '';
+    };
+
+    flags = {
+      type = types.listOf types.string;
+      description = ''
+        Flags to be automatically appended when running nnn.
+
+        Disjoint with the `configFile` option.
```
llakala: Is it, or is this a typo?

modules/nnn.nix
```diff
@@ -0,0 +1,77 @@
...
+      if hasFlags then
+        { inherit (options) flags; }
+      else
+        {}
+    )
+    // (
+      if hasConfig then
+        {
+          symlinks."$out/share/nnn/nnn.rc" =
+            options.configFile or (writeText "nnn.rc" options.configContents);
```
llakala: Nit - we use `if/else if / else null` in cases like this. It's identical behavior here, but it's standard practice. See my example code on the merge chain. We also always expand out `symlinks.foo` and `environment.foo`.

modules/nnn.nix
```diff
@@ -0,0 +1,77 @@
...
+          environment.XDG_CONFIG_HOME = "$out/share";
```
llakala: The wrapped symlink goes at toplevel, not under the `share` dir.

modules/nnn.nix
```diff
@@ -0,0 +1,77 @@
...
+    package = {
+      type = types.derivation;
+      defaultFunc = { inputs }: inputs.nixpkgs.pkgs.nnn;
+      description = "The nnn package to be wrapped.";
+    };
+  };
+
+  impl =
+    { options, inputs }:
+    assert !(options ? configContents && options ? configFile);
```
llakala: Nit - assertions go _below_ the `let/in` statement.

modules/nnn.nix
```diff
@@ -0,0 +1,77 @@
...
+    let
+      inherit (inputs.nixpkgs.pkgs) writeText;
+      hasFlags = options ? flags;
+      hasConfig = options ? configFile || options ? configContents;
+    in
+    inputs.mkWrapper {
+      inherit (options) package;
+    }
+    // (
```
llakala: We don't need a whole merge chain here.
```nix
inputs.mkWrapper {
  inherit (options) package;
  flags = options.flags or [];
  symlinks = {
    "$out/nnn/nnn.rc" =
      if options ? configFile then
        options.configFile
      else if options ? configContents then
        writeText "nnn.rc" options.configContents
      else
        null;
  };
  environment = {
    XDG_CONFIG_HOME = "$out";
  };
}
```
We support `null` for symlinks and env vars precisely for this usecase (and empty flags are treated as a no-op). This doesn't set `environment` conditionally, but that would only matter if the user didn't set a `configFile` OR `configContents`, and having XDG_CONFIG_HOME set somewhere else wouldn't matter.

(Also nit while I'm here, if the program supports a flag instead of overriding xdg, we prefer that.)

llakala (changes requested):

modules/nnn.nix
yaaaarn: typo

modules/nnn.nix
```diff
@@ -0,0 +1,72 @@
...
+        Disjoint with the `configContents` option.
+      '';
+    };
+
+    flags = {
+      type = types.listOf types.string;
+      description = ''
+        Flags to be automatically appended when running nnn.
+
+        Disjoint with the `configContents` option.
```
llakala: You've introduced a new typo instead? Isn't it just not disjoint with anything?

yaaaarn: silly me 😭

llakala (approved):

